Issue #19, opened 2021-11-26
Implement sign up and sign in with Google and Apple


cacarrara:
## Description
Terraso will support only sign up and sign in based on 3rd party services (Google and Apple only). To accomplish that, we will have to implement on Terraso the support for OAuth and (eventually) OpenID.

## Resources
- [Authenticating users with Sign in with Apple](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple)
- [Authenticate with a back-end server - Google](https://developers.google.com/identity/sign-in/web/backend-auth)
- [Django OAuth toolkit](https://django-oauth-toolkit.readthedocs.io/en/latest/) (only a suggestion)

## Suggested subtasks
- [x] Study Apple and Google documentations and design a solution for integrated authentication. It's important to consider the front-end → back-end Terraso flows here.
- [x] Implement support for Sign up / Sign in with Google
- [x] Implement support for Sign up / Sign in with Apple

PR #48, opened 2021-12-14
feat: adds authentication feature

cacarrara:
This change makes necessary the user be authenticated to perform GraphQL
queries.

To use GraphiQL user interface, the user have to be logged in to Django
admin (initially this will be only available for admins).

To use GraphQL API, the requests must have a valid JWT token on
Authentication header.

The JWT token (and refresh token) is informed on response of 
`POST /auth/google/authorize` among other user data. Example:

```json
{
  "user": {
    "email": "user@gmail.com",
    "first_name": "Name",
    "last_name": "Lastname",
    "access_token": "jwt_token_a",
    "refresh_token": "jwt_token_r"
  }
}
```
The JWT token has a default expiration time of 6 minutes and the refresh token
has expiration time of 1 hour. Both deltas are configurable by env vars.

Once the access token is expired, a new pair of tokens (access and refresh)
can be requested on `POST /auth/tokens` informing a previous valid refresh
token on request body (JSON).

This is part of:
- #19


cacarrara:
I changed a bit the approach of adding this feature to don't require a break deployment. So, I'll close this PR and open another one with proper explanation.

PR #58, opened 2021-12-22
feat: add authentication callback for Google and Apple

cacarrara:
This change adds the endpoint responsible for processing the Apple
and Google callback after user authorize Terraso application access 
their data. By the end of the success process, the API procedure will
redirect user back to the web client application with auth details 
attached in a cookie.

It's important to mention that the cookie value isn't still a JWT because
this part depends on integrating this PR with the following one:

- #48 

I intend to do that in a separated PR.

### DIFF
--- a/terraso_backend/config/settings.py
+++ b/terraso_backend/config/settings.py
@@ -113,10 +113,11 @@ GRAPHENE = {
     "SCHEMA": "apps.graphql.schema.schema",
 }
 
+WEB_CLIENT_URL = config("WEB_CLIENT_ENDPOINT", default="")
+AUTH_COOKIE_DOMAIN = config("AUTH_COOKIE_DOMAIN", default="terraso.org")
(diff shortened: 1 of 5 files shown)


terraso_backend/config/settings.py
```diff
@@ -113,11 +113,12 @@
     "SCHEMA": "apps.graphql.schema.schema",
 }
 
+WEB_CLIENT_URL = config("WEB_CLIENT_ENDPOINT", default="")
+AUTH_COOKIE_DOMAIN = config("AUTH_COOKIE_DOMAIN", default="terraso.org")
```
josebui, line 117: Question: Should this contain the subdomain too?

josebui (approved):

terraso_backend/config/settings.py
cacarrara: This is going to consider all subdomains.

PR #62, opened 2021-12-23
fix: returns 401 status code for unauthenticated requests

cacarrara:
This change updates the HTTP response status code for unauthenticated
requests. Before it was returning 403 (Forbidden) while 401
(Unauthorized) is more appropriate.

### DIFF
--- a/terraso_backend/apps/auth/views.py
+++ b/terraso_backend/apps/auth/views.py
@@ -128,7 +128,7 @@ class CheckUserView(View):
     def get(self, request, *args, **kwargs):
         if not request.user.is_authenticated:
             return JsonResponse(
-                {"error": "Permission denied"}, status=403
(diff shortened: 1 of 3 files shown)


josebui (approved):

terraso_backend/apps/auth/views.py
```diff
@@ -128,7 +128,7 @@ class CheckUserView(View):
     def get(self, request, *args, **kwargs):
         if not request.user.is_authenticated:
             return JsonResponse(
-                {"error": "Permission denied"}, status=403
+                {"error": "Unauthenticated request."}, status=401
```
paulschreiber, line 131: `Unauthorized` is the standard message here.

paulschreiber (changes requested):

PR #64, opened 2021-12-24
fix: gracefully handle nth apple login

paulschreiber:
## Description
Since first and last names are not supplied by Apple on logins after the first, don't attempt to use null/empty fields or insert them in to the database.

Resolves the following errors:
```
Error: couldn't get User name from Apple
```

```
Error: null value in column "first_name" of relation "core_user" violates not-null constraint
DETAIL: Failing row contains (, null, f, null, null, f, t, 2021-12-24 02:58:45.630905+00,
6323b526-d75d-43fd-bc0d-eeb2d66422ba, 2021-12-24 02:58:45.631401+00, 2021-12-24 03:19:11.411189+00, 
xxxx@gmail.com).
```

### Verification steps
1. Sign in to Terraso with Apple
2. Log out
3. Sign in to Terraso with Apple

### DIFF
--- a/terraso_backend/apps/auth/services.py
+++ b/terraso_backend/apps/auth/services.py
@@ -36,12 +36,16 @@ class AccountService:
         if not tokens.is_valid:
             raise Exception("Error fetching auth tokens: " + tokens.error_description)
 
+        name_defaults = {}
+        if first_name:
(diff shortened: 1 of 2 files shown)


cacarrara:
This was fixed by:
- #66

Issue #22, opened 2021-11-26
Add photo field to the User model


cacarrara:
## Description
The user profile photo might be automatically fetched from the third-party account system (Google or Apple), or it can also be uploaded from by the user. Since the file itself might be stored on an external storage service, this field will be used to store the location of the file.

In this issue, it's important to consider the flow front-end → back-end for photo upload.

## Suggested subtasks
- [ ] Design the overall flow to upload photo considering front-end → back-end flow
- [ ] Add the new field on model with proper support to the external storage service (upload) and update DB migrations
- [ ] Implement upload feature to update photo
- [ ] Add support to present the proper photo URL from external services
- [ ] Add the new photo field on user API

This issue depends on:
- #21

PR #74, opened 2022-01-06
feat: User profile image

### LINKED ISSUE #21: Add support for 3rd party file storages (AWS S3, for example)
## Description
It's expected from Terraso the store of several files like user profile photo, Landscape map files and other in the future. To enable these features, Terraso will have to integrate to third-party storage services. At the beginning, we should be integrating with AWS S3.

This issue refers only to make the appropriate changes to the project to make it possible the usage of the storage services. The features that use the storage itself will be specified on other issues.

## Resources
- [django-storages](https://django-storages.readthedocs.io/en/latest/index.html) (suggestion)

josebui:
## Description

- Added S3 integration
- Added google profile image update
- Added endpoint to upload profile image

### Checklist
- [x] New tests added

### Related Issues
Fixes #21 #22


cacarrara (commented):
@josebui could you please apply `black` formatter to the code being added? We'll have to put this check on lint later. :sweat_smile:

terraso_backend/config/settings.py
```diff
@@ -139,3 +139,5 @@
     "JWT_REFRESH_EXP_DELTA_SECONDS", default="3600", cast=config.eval
 )
 JWT_ISS = config("JWT_ISS", default="https://terraso.org")
+
+PROFILE_IMAGES_S3_BUCKET = config("PROFILE_IMAGES_S3_BUCKET")
```
cacarrara, line 143: I think it would be better to add the AWS credentials to the settings file as well. Let it be only on environment variables leaves it a bit over implicit. `django-storages` describes the settings here: https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#settings

terraso_backend/apps/storage/services.py
```diff
@@ -0,0 +1,11 @@
+from .s3 import ProfileImageStorage
+
+
+class ProfileImageService:
+    def upload(self, user_id, image):
+        profile_image_storage = ProfileImageStorage()
+        profile_image_storage.save(user_id, image)
```
cacarrara, line 7: Doesn't this call raise any exception? What happens when there's a connectivity issue, for example?

terraso_backend/apps/auth/services.py
```diff
@@ -48,8 +56,22 @@ def _persist_user(self, email, first_name="", last_name=""):
...
 
+        user_id = str(user.id)
+        has_picture = profile_image_service.exists(user_id)
+        if not has_picture and picture_file:
+            image_file = ContentFile(picture_file)
+            profile_image_service.upload(user_id, image_file)
+
         return user
 
+    def _get_profile_image(self, url):
```
cacarrara, line 67: Can we move all this file handling to the storage app? We could have something like `ProfileImaageService().upload_from_url(user_id, remote_url)`. I miss some exceptions handling, and it's going to be better to hide these complexities under storage app. Auth shouldn't be dealing with files.

cacarrara (changes requested):
I missed the place where the URL from the uploaded image is persisted to the user model (database) and this information being returned on API to the web client.

josebui:
Hey @cacarrara :
- About `black` can you point me to a document that describes how to do that? I don0t have experience with that
- About persisting the image to the user model: There is no need (functionally) to do that right now, since the S3 filename will be the user ID so there is no need to do this to provide the URLs to the web client

terraso_backend/apps/storage/services.py
josebui: Good point, I will check that

terraso_backend/apps/auth/services.py
josebui: @cacarrara This is a particular process related to how google is sending the picture data, this is not related to the `storage` app, I think it is better to leave this here related to the auth providers, maybe I can change the name to `_get_google_profile_image`

cacarrara: I'm not referring to the specific process of taking a Google image. I'm referring to the generic process of getting a file from a remote location and putting into our storage. I really don't think it's appropriate to make file handling on authentication app. This breaks the single responsibility principle.

cacarrara:
> About persisting the image to the user model: There is no need (functionally) to do that right now, since the S3 filename will be the user ID so there is no need to do this to provide the URLs to the web client

How web-client is going to know the user profile URL? Anyway, it's important to have this data on our database regardless of anything else. This is a user's information that must be persisted on user's table. What if a mobile client needs to show the user's picture...

> About black can you point me to a document that describes how to do that? I don0t have experience with that

You can configure your editor to run black. Or, you can simply do:

```sh
pip install black
black .
```

More info here: https://github.com/psf/black

josebui:
@cacarrara: 
**How web-client is going to know the user profile URL?**
The web client will get the URL like: `https://{bucket-name}.s3.us-east-2.amazonaws.com/{user-id}`

**Anyway, it's important to have this data on our database regardless of anything else.**
Since no  product requirement is driving this, can you add more context on why this is "important"? related to the current product goals. The requirement is to provide profile images, and with only this changes that is being provided.

**This is a user's information that must be persisted on user's table.**
Same thing, why we `must` do this is not linked to product objectives, as far as I can see, if it eventually is something we need to provide at that level we can add it.

**What if a mobile client needs to show the user's picture...**
The same way as any client: `https://{bucket-name}.s3.us-east-2.amazonaws.com/{user-id}`

cacarrara:
> ... any client https://{bucket-name}.s3.us-east-2.amazonaws.com/{user-id}

This is duplication of information and, despite some specific exceptions, it's a bad decision by itself, and we should avoid as much as possible. Both API and any other client (even those we eventually won't control) will have to know the bucket name, to know that we use s3 as storage, in which region it's stored and so on... This is bad because:

1. We make information scattered so;
2. We compromise maintainability. What if we start using a CDN that has dynamic URLs for different users? We would have to change every client. Keeping the URL only persisted on back-end not only avoid this maintenance burden, but we also make the system more flexible;

This is not a product matter, it's a software / system matter. I don't know if I understood correctly, but are you expecting product specifications saying we should persist something in database? I'm a bit confused here.

terraso_backend/apps/auth/services.py
josebui: That makes more sense to me, I think that can fall under the storage responsabilities because it can upload a url resrouce to a file, I can change that

josebui:
FYI to anyone reading this PR, the conversation continued through direct messages, just to add my side of this, this is the value I see on the approach I took:
- Less code to maintain
- Avoid holding data unnecessarily in the DB
- Minimal changes that can provide value without affecting the product

terraso_backend/apps/storage/services.py
```diff
@@ -0,0 +1,30 @@
...
+        self.profile_image_storage.save(self._get_image_path(user_id), ContentFile(image))
+
+    def upload_file(self, user_id, image):
+        self.profile_image_storage.save(self._get_image_path(user_id), image)
+
+    def exists(self, user_id):
+        return self.profile_image_storage.exists(self._get_image_path(user_id))
+
+    def get_url(self, user_id):
+        return "{}/{}/profile-image".format(settings.PROFILE_IMAGES_BASE_URL, user_id)
```
paulschreiber, line 23: Have this use `_get_image_path()`

terraso_backend/apps/auth/services.py
```diff
@@ -48,6 +54,14 @@ def _persist_user(self, email, first_name="", last_name=""):
         if update_name:
             user.save()
 
+        user_id = str(user.id)
+        try:
+            has_picture = profile_image_service.exists(user_id)
+            if not has_picture and profile_image_url:
+                profile_image_service.upload_url(user_id, profile_image_url)
+        except Exception:
+            logger.exception("Failed to upload profile image")
```
paulschreiber, line 63: We don't log exceptions anywhere else. If we are going to log things, it would be good to do so (a) with rollbar and (b) more broadly. In this specific case, we should include the user ID as part of the message.

For now, have this return `None`. I opened a new ticket to handle error logging.

paulschreiber (changes requested):
How does this handle checking if the user has an image? Are we calling an S3 API every time? That is expensive. Another advantage of persisting the data is we can quickly check how many users have images and if a particular user has an image.

josebui:
@paulschreiber That is a good point, I didn't know AWS charged for not found s3 objects requests, but that seems to be the case. I will persist this data to the DB

terraso_backend/apps/auth/views.py
```diff
@@ -133,15 +135,18 @@ def post(self, request, *args, **kwargs):
...
         return JsonResponse(
             {
                 "user": {
-                    "email": request.user.email,
-                    "first_name": request.user.first_name,
-                    "last_name": request.user.last_name,
+                    "email": user.email,
+                    "first_name": user.first_name,
+                    "last_name": user.last_name,
+                    "profile_image": profile_image_service.get_image_url(user.profile_image_path),
```
cacarrara, line 149: The `user.profile_image_path` hasn't already the built image URL? Why to call the service method here?

terraso_backend/apps/core/models/users.py
```diff
@@ -47,6 +47,7 @@ class User(AbstractUser):
 
     username = None
     email = models.EmailField(unique=True)
+    profile_image_path = models.CharField(max_length=128, null=True)
```
cacarrara, line 50: We should use a `URLField` here among other things, it's going to have a more reasonable length, for example.

terraso_backend/apps/graphql/schema/users.py
```diff
@@ -4,11 +4,18 @@
 from graphene_django import DjangoObjectType
 
 from apps.core.models import User
+from apps.storage.services import ProfileImageService
 
 from .commons import BaseDeleteMutation
 
 
 class UserNode(DjangoObjectType):
+    profile_image = graphene.String(description="The profile image URL")
```
cacarrara, line 13: We shouldn't need that if the data persisted to the database is the complete / built image url. It's just a matter of adding the model field to the `Meta.fields` attribute.

cacarrara (changes requested):

terraso_backend/apps/auth/views.py
josebui: @cacarrara To avoid being tightly coupled to S3 or the the domain from where we can access the images, the only thing we will be stored in the DB is the path to the image inside the source of truth of the files. That is why we need to create the complete URL here

terraso_backend/apps/core/models/users.py
josebui: @cacarrara It will not be the complete URL, see: https://github.com/techmatters/terraso-backend/pull/74#discussion_r782353799

terraso_backend/apps/graphql/schema/users.py
josebui: @cacarrara Same thing, the DB field will not contain the complete URL. See: https://github.com/techmatters/terraso-backend/pull/74#discussion_r782353799

terraso_backend/apps/core/models/users.py
josebui: @cacarrara Let me know if you have a suggestion for the length of the path, 128 felt reasonable, but let me know

terraso_backend/apps/auth/views.py
cacarrara: I was thinking to persist the entire URL exactly to be decoupled from any specific storage. Once we uploaded the file and got the URL at the upload moment, we don't need to care about the source of the truth of the files anymore. Let's suppose we start working with S3 now and at some point we decide to start uploading image to `imgur` (absurd example). Then, with this implementation, we got stuck to S3, or we have to work around on application level to build the proper URL for the other service.

Despite the point above, let's suppose we have to dump the DB or export the DB data to any format, like CSV for any reason. Then, we wouldn't have the complete profile image URL as well, unless we pass through the application layer.

What do you think about that?

paulschreiber: I agree that we should store the full URL, for the reasons Caio states.

josebui: I see, I didn't think about those potential scenarios, so lets gather all of our concerns:
1. We might want to change the current storage for user images
2. We might want to use more than one storage location for user images
3. The DB doesn't have the full location of the user images, so we have to always relay on the application layer to identify the complete URL.

Lets do a simplistic "risk" analysis exercise of the potential scenarios if those arise in the future with the 2 approaches:

Effort: (0-5(high)) Effort to change the implementation to support this
Likelihood: (0-5(high)) Likelihood that the future scenario will appear

**Current Implementation: Persist only image path** (value: 38)

1. We might want to change the current storage for user images (value: 6)
  Effort: 1 (Change the constants, move files to new location)
  Likelihood: 6 (Having in mind that we don't have our infrastructure in AWS right now it is likely that we don't want to use S3 in the future)
2. We might want to use more than one storage location for user images (value: 8)
  Effort: 4 (Create script to complete URL appending the new base path, add support for extra storage)
  Likelihood: 2 (Seems that this scenario will be really unlikely, I don't se a reason to support multiple storages at the top of my mind)
3. DB is not the source of truth for files URLs, and we need to gather data from there directly (value:  24)
  Effort: 3 (Besides the script to generate the data we need an extra step to complete the URLs)
  Likelihood: 8 (It seems really possible that we might want to extract data directly from the DB)
  
**Potential implementation: Persist complete URL** (value: 28)

1. We might want to change the current storage for user images (value: 24)
  Effort: 4 (Create a script to gather all persisted URLs change them and save them again with the new URL, move files to the new location)
  Likelihood: 6 (Having in mind that we don't have our infrastructure in AWS right now it is likely that we don't want to use S3 in the future)
2. We might want to use more than one storage location for user images (value: 4)
  Effort: 2 (just add support for extra storage)
  Likelihood: 2 (Seems that this scenario will be really unlikely, I don't se a reason to support multiple storages at the top of my mind)
3. DB is not the source of truth for files URLs, and we need to gather data from there directly (value: 0)
  Effort: 0 (Already provided by the solution)
  Likelihood: 8 (It seems really possible that we might want to extract data directly from the DB)
  
@cacarrara It seems that your approach to hold the full URL has less potential risk, but let me know if you see a whole in my logic haha

cacarrara: Wow. You went really deep into it :astonished: -- Congrats and thanks for this analysis. My thought always was more about respecting some principles (don't repeat yourself, separation of concerns, single responsibility). Also, keeping the application flexible enough for changes. We cannot predict all possible changes we can have (thanks for the exercise of likelihood). While probably we won't have more than one storage for images, things like migrate the current schema to support CDNs, or even some sort of migration where we can incrementally migrate the DB, seems reasonable. I think having the URL on DB enables this kind of flexibility, respecting the principles.

terraso_backend/apps/core/models/users.py
```diff
@@ -47,6 +47,7 @@ class User(AbstractUser):
 
     username = None
     email = models.EmailField(unique=True)
+    profile_image = models.URLField(null=True)
```
cacarrara, line 50: It's a detail, but to avoid having two values for "no value" (null and empty string), it's a common practice to define the field this way: `models.URLField(blank=True, default="")`. It means the model allows no value for this field and the default value (when nothing is informed) is going to be and empty string. This is how `Group.website` and `Landscape.website` is being declared.

Remember to generate the migration again after the change.

cacarrara (approved):

PR #82, opened 2022-01-12
build: check if docker is running before calling docker-compose up

paulschreiber:
## Description
Check if Docker is running before calling `docker-compose` up. Avoids long, messy errors like so:
```
$ make run
./scripts/rebuild.sh
Checking if we need a docker images rebuild before running this command
docker build --tag=techmatters/terraso_backend --file=Dockerfile .
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
ERRO[0000] Can't add file /Users/paul/.dev/backend/.git/hooks/pre-push.sample to tar: io: read/write on closed pipe 
make[1]: *** [build_base_image] Error 1
make: *** [check_rebuild] Error 2
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,8 +48,8 @@ setup-git-hooks:
 
 pre-commit: lint
 
-run: check_rebuild
-	@docker-compose up
+run:
+	@./scripts/docker.sh
 
 setup: build setup-pre-commit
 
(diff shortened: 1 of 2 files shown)


cacarrara (commented):
We also have other commands that depend on docker running: `build_base_image`, `build`, `start`, `stop` and so on. What about these other commands?

cacarrara (approved):
I don't agree with this approach. Mainly because I think it's not the responsibility of the application to do this kind of work. 

But, if it's something important for you and, considering that I cannot see bigger practical implications right now other than "conceptual" problem, I'll approve it.

Issue #70, opened 2022-01-04
Group management API changes


paulschreiber:
## Description

Changes needed:
- [x] Make the Group creator automatically have the Manager role;
- [x] Allow only Group Managers to update their groups;
- [x] Create / adapt API to Group Managers remove members;
- [x] Do not allow empty Groups. A Group always need to have at least one Manager;

Web client issue: https://github.com/techmatters/terraso-web-client/issues/97

cacarrara:
Regarding:

> Create / adapt API to Group Managers remove members;

We already have a GraphQL mutation to perform this operation: `deleteMembership()`. The PR #95 adds the proper permission checks before deleting a `Membership`.

PR #95, opened 2022-01-19
feat: creates permission to delete Membership

cacarrara:
This change creates permission to a given User to delete a Membership. The
implemented permission follows the rules:

- A User can always delete their own Memberships;
- A Group Manager can always delete any Memberships from their managed
  Groups;
- A Group always need to have at least one Manager.

This is part of:
- #70


terraso_backend/apps/core/permission_rules.py
```diff
@@ -4,3 +4,39 @@
...
+    is_own_membership = user.memberships.filter(pk=membership_id).exists()
+
+    # User is the last manager and a Group cannot have no managers
+    if managers_count == 1 and is_user_manager and is_own_membership:
+        return False
+
+    return True
+
+
+rules.add_rule("allowed_group_managers_count", allowed_group_managers_count)
```
josebui, line 42: Question: How does this work `rules.add_rule(`? it seems the only rule using this is `allowed_group_managers_count`why is that?

josebui (approved):
Just one question, it looks good

paulschreiber (approved):

terraso_backend/apps/core/permission_rules.py
cacarrara: Good question, @josebui. 

We have 3 predicates in this file, and we use them in two different ways. The first 2 predicates are associated directly to the model permissions (add, change, delete). They are referenced on `Meta` class on proper models. The third predicate, added in this PR, is added to the `rules` rule set. 

To evaluate a rule on rule set, we can call it as `rules.test_rule('rule_name', params)`.

Some examples that might help understand it:
```python
>>> @rules.predicate
>>> def is_book_author(user, book):
...     return book.author == user
>>>
>>> rules.add_rule('can_edit_book', is_book_author)
>>> rules.test_rule('can_edit_book', adrian, guidetodjango)
```

Using it with Django models:
```python
class Book(RulesModel):
    class Meta:
        rules_permissions = {
            "add": rules.is_staff,
            "read": rules.is_authenticated,
        }

# This is equivalent to:
rules.add_perm("app_label.add_book", rules.is_staff)
rules.add_perm("app_label.read_book", rules.is_authenticated)
```

josebui: Great, thanks for the explanation Caio

Issue #69, opened 2022-01-04
Add API call to delete groups and landscapes


paulschreiber:
## Description
Add API call to delete groups and landscapes
- [x] Ensure user requesting deletion is owner
- [x] Perform soft delete (mark as deleted)
- [x] Filter list and get calls to exclude soft-deleted items

PR #100, opened 2022-01-21
feat: permissions to delete / change Group and Landscape

cacarrara:
This change adds:

- Permission checks to delete Landscape;
- Permission checks to update Landscape;
- Permission checks to delete Group;

It implements:
- #69

### DIFF
--- a/terraso_backend/apps/graphql/schema/groups.py
+++ b/terraso_backend/apps/graphql/schema/groups.py
@@ -95,3 +95,15 @@ class GroupDeleteMutation(BaseDeleteMutation):
 
     class Input:
         id = graphene.ID()
+
+    @classmethod
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        user = info.context.user
+
+        ff_check_permission_on = settings.FEATURE_FLAGS["CHECK_PERMISSIONS"]
+        user_has_delete_permission = user.has_perm(Group.get_perm("delete"), obj=kwargs["id"])
+
+        if ff_check_permission_on and not user_has_delete_permission:
+            raise GraphQLValidationException("User has no permission to delete this data.")
+
+        return super().mutate_and_get_payload(root, info, **kwargs)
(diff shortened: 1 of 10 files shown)


terraso_backend/apps/graphql/schema/groups.py
```diff
@@ -95,3 +95,15 @@ class GroupDeleteMutation(BaseDeleteMutation):
 
     class Input:
         id = graphene.ID()
+
+    @classmethod
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        user = info.context.user
+
+        ff_check_permission_on = settings.FEATURE_FLAGS["CHECK_PERMISSIONS"]
```
paulschreiber, line 103: Can we remove the feature flag and assume a concurrent deploy of the frontend work?

terraso_backend/apps/graphql/schema/groups.py
```diff
@@ -95,3 +95,15 @@ class GroupDeleteMutation(BaseDeleteMutation):
...
+        user_has_delete_permission = user.has_perm(Group.get_perm("delete"), obj=kwargs["id"])
+
+        if ff_check_permission_on and not user_has_delete_permission:
+            raise GraphQLValidationException("User has no permission to delete this data.")
```
paulschreiber, line 107: Will this exception ever be shown to the user? i.e. do we need to localize it?

cacarrara: Yes. It might be shown to the user, and we will have to localize. I've been thinking in ways to keep localization only on web client. But I don't think it's the right way to do that. Perhaps we should have localization on API as well.

terraso_backend/apps/graphql/schema/groups.py
cacarrara: I would rather keep the feature flag until we make the proper changes on web client.

terraso_backend/apps/graphql/schema/groups.py
paulschreiber: Can we return something like `error.permission.delete` or so?

terraso_backend/apps/graphql/schema/groups.py
josebui: Thought: I like feature flags, I think the important process here will be how to maintain it later, maybe at some point it is not needed and we should remove the flag. For this particular situation since a coordination is required between client and api I think it makes sense to keep it

josebui (approved):

PR #101, opened 2022-01-25
fix: GraphQL Group filters using deleted associated objects

cacarrara:
There was a specific case where using Django built-in lookups for
relationships don't work properly with the soft delete solution we
implemented with `django-safedelete`. According to the library
maintainers, it's a known limitation[1].

To work around this limitation, a customized filter set were created for
Group schema on GraphQL. This custom filter set implements the
adjustments expected to ignore soft deleted objects when filtering by
relationships (joins).

[1] - https://github.com/makinacorpus/django-safedelete/issues/158

### DIFF
--- a/terraso_backend/apps/graphql/schema/groups.py
+++ b/terraso_backend/apps/graphql/schema/groups.py
@@ -1,3 +1,4 @@
+import django_filters
 import graphene
 from django.conf import settings
 from graphene import relay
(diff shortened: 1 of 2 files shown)


terraso_backend/apps/graphql/schema/groups.py
```diff
@@ -9,24 +10,40 @@
 from .commons import BaseDeleteMutation, BaseWriteMutation, TerrasoConnection
 
 
-class GroupNode(DjangoObjectType):
-    id = graphene.ID(source="pk", required=True)
+class GroupFilterSet(django_filters.FilterSet):
+    # TODO: members__email was kept for backward compatibility. Remove as soon
+    # as the web client be updated to use memberships__email filter
```
paulschreiber, line 15: question: do we have an issue tracking this? If not, please make one.

paulschreiber (approved):

terraso_backend/apps/graphql/schema/groups.py
cacarrara: Created:
- https://github.com/techmatters/terraso-web-client/issues/182
- https://github.com/techmatters/terraso-backend/issues/105

PR #106, opened 2022-01-26
feat: removes LandscapeGroupUpdateMutation

cacarrara:
Landscape Group (the relationship model between Landscape and Group)
isn't expected to be updated. The only attribute this entity has is the
`is_default_landscape_group` that is a more internal attribute and won't
be updated.

### DIFF
--- a/terraso_backend/apps/graphql/schema/__init__.py
+++ b/terraso_backend/apps/graphql/schema/__init__.py
@@ -17,7 +17,6 @@ from .landscape_groups import (
     LandscapeGroupAddMutation,
     LandscapeGroupDeleteMutation,
     LandscapeGroupNode,
-    LandscapeGroupUpdateMutation,
 )
 from .landscapes import (
     LandscapeAddMutation,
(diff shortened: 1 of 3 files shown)


josebui:
@cacarrara Question: I think I am still al ittle confused about this concept, it seems that there are 3 types of groups:
- Group (not attached to landscapes?)
- Landscape group (not default group?)
- Landscape default group

Is this the case? Or is the `landscape default group` a special case of a `landscape group`?

cacarrara:
@josebui I think you are right in your assumptions. As an attempt to clarify a bit more the model's structure, check the diagram below:
```                           
      +----------------+   
      |    Landscape   |   
      +----------------+   
               |           
               |           
               v           
     +-------------------+ 
     |  LandscapeGroup   | 
     +-------------------+ 
              ^            
              |            
              |            
     +------------------+  
     |       Group      |  
     +------------------+  
```

The `LandscapeGroup` is just an intermediary model between `Landscape` and `Group`. It has one attribute that says if that relationship indicates that the group is the default one for that `Landscape` or not. The default Group only exists to make possible users to "join a Landscape".

So, long story short, we have only one type of Group that might have two types of relationships with a `Landscape`: default or non default. The default `Group` of a `Landscape` soon will be created automatically during the `Landscape` creation. I think the default Group also doesn't need to be visible to the users, and perhaps we will have to properly filter them out on API queries. But this is another topic... :sweat_smile:

paulschreiber (approved):

Issue #75, opened 2022-01-08
Audit all API calls for permission checks


paulschreiber:
## Description
Ensure all calls (such as leaving a group) check permissions before executing.

paulschreiber:
This looks good. To clarify: _everybody_ still requires authentication (i.e. all logged in users, not everybody).

cacarrara:
> This looks good. To clarify: _everybody_ still requires authentication (i.e. all logged in users, not everybody).

Sure.

PR #107, opened 2022-01-26
feat: adds permission checks to mutate Landscape Group

cacarrara:
This change adds permission check to allow only Landscape managers to
add Landscape Group. It means only a Landscape Manager can create a new
association to a Group.

It also adds permission check to allow only the Landscape or Group
manager to delete a Landscape Group relationship.

This is part of:
- #75

### DIFF
--- a/terraso_backend/apps/graphql/schema/landscape_groups.py
+++ b/terraso_backend/apps/graphql/schema/landscape_groups.py
@@ -47,9 +47,9 @@ class LandscapeGroupAddMutation(relay.ClientIDMutation):
         group = Group.objects.get(slug=kwargs.pop("group_slug"))
 
         ff_check_permission_on = settings.FEATURE_FLAGS["CHECK_PERMISSIONS"]
-        user_has_delete_permission = user.has_perm(LandscapeGroup.get_perm("add"), obj=landscape.pk)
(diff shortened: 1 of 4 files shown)


terraso_backend/apps/graphql/schema/landscape_groups.py
```diff
@@ -41,9 +42,19 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
         called both when adding and updating Landscape Groups. The `kwargs`
         receives a dictionary with all inputs informed.
         """
+        user = info.context.user
+        landscape = Landscape.objects.get(slug=kwargs.pop("landscape_slug"))
+        group = Group.objects.get(slug=kwargs.pop("group_slug"))
+
+        ff_check_permission_on = settings.FEATURE_FLAGS["CHECK_PERMISSIONS"]
+        user_has_delete_permission = user.has_perm(LandscapeGroup.get_perm("add"), obj=landscape.pk)
```
josebui, line 50: suggestion(non-blocking): Maybe the `user_has_delete_permission` only needs to exist if the feature flag is true to avoid that computation if the feature is not enabled

josebui (approved):
Looks good, just one suggestion

paulschreiber (approved):

PR #115, opened 2022-01-28
feat: add error codes to validation error responses

cacarrara:
This change updates how the validation messages are built. It now makes
the message be a serialized JSON in order to add more contextual data to
the message. This is an attempt to make possible the i18n of validation
messages by the clients.

This is an example of validation error message using the new schema:

```
'message': '[{
    "code": "updateNotAllowed",
    "context": {"model": null, "field": "membership"}
}]'
```

If the validation happens in a specific field, the context `field`
attribute will have the related field name. Otherwise, it will be null. 

```
'message': '[{
    "code": "notFound",
    "context": {"model": "Membership", "field": "group", "extra": ""}
}]'
```

In some specific cases, an error message might have a complement to the
error code. Like in a situation where a `Membership` cannot be updated
because the user is the last manager in that group:

```
'message': '[{
    "code": "updateNotAllowed",
    "context": {"model": "Membership", "field": null, "extra": "manager_count"}
}]'
```

In cases where more than one field have validation errors, then the
message will have more than one object:

```
'message': '[
    {"code": "blank", "context": {"model": "Landscape", "field": "name"}},
    {"code": "invalid", "context": {"model": "Landscape", "field": "website"}
}]'
```

To make this structure work, more specific exceptions were added to the
GraphQL app with the proper handler to format the error message. Also,
all mutations were already updated to work with the new schema.

### DIFF
--- a/terraso_backend/apps/graphql/exceptions.py
+++ b/terraso_backend/apps/graphql/exceptions.py
@@ -5,8 +5,24 @@ class GraphQLValidationException(Exception):
 
         for field, validation_errors in validation_error.error_dict.items():
             for error in validation_errors:
-                messages.extend([f"field={field}, error={message}" for message in error.messages])
+                messages.append(_format_error_message(field, error.code))
 
-        error_message = ";".join(messages)
+        error_message = "; ".join(messages)
 
         return cls(error_message)
+
+
+class GraphQLNotFoundException(GraphQLValidationException):
+    def __init__(self, message="", field=None):
+        super().__init__(_format_error_message(field, "not_found", extra=message))
+
+
+class GraphQLNotAllowedException(GraphQLValidationException):
+    def __init__(self, message="", field=None, operation="operation"):
+        super().__init__(_format_error_message(field, f"{operation}_not_allowed", extra=message))
+
+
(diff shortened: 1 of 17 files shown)


paulschreiber (approved):

josebui:
@cacarrara Some comments:
- Suggestion: It makes sense to me that the context of the error might change depending on the procces, I think in general for validation this applies but maybe there are other process that will not have this structure, from that I think having something like `{ "errorCode": "", context: { fieldName: "", entity: "", ...} }` might be more flexible and easier to map in the client side
- Suggestion: Is it possible to return this data in a json format? I think that will make the parsing much easier, and also make this logic easier to extend. `[{ "errorCode": "", context: { fieldName: "", entity: "", ...} }, ...]`
- Suggestion: Also, graphQL is returning the variable as cammel cased not snake cased, I think it is a good idea to keep it consistent for the clients

cacarrara:
Thanks, @josebui. I'll work on the suggestions :+1:

cacarrara:
@josebui I changed the message to be a JSON. Take a look and tell me what you think :slightly_smiling_face:

PR #122, opened 2022-02-03
feat: add command to load Landscapes' GeoJSON boundaries

cacarrara:
This change adds a custom command to load Landscapes' polygon area
directly from Airtable.

Fix: #121


paulschreiber (approved):
question: can we rename loadlandscapegeojson to have underscores in the file name without affecting the command line argumentt?

PR #133, opened 2022-02-15
feat: User preferences

### LINKED ISSUE #131: Remember user's language choice
## Description
For each user, store the most recently selected language (i.e. `en` or `es`) in the database, and use that language upon login.

## Additional context

Dependant issue: https://github.com/techmatters/terraso-web-client/issues/185

josebui:
## Description

- Added UserPreference model
- Added update preference mutation

### Checklist
- [ ] New tests added

### Related Issues
Fixes #131


terraso_backend/apps/core/models/users.py
```diff
@@ -81,3 +83,19 @@ def is_group_manager(self, group_id):
 
     def __str__(self):
         return self.email
+
+
+class UserPreference(BaseModel):
+    key = models.CharField(max_length=128)
+    value = models.TextField(max_length=512, blank=True, default="")
```
cacarrara, line 90: Question: why text field instead of char field?

cacarrara, line 90: Suggestion: checking the structure now of preferences, what do you think of adding a `JSONField` on `User` model? We have a field like that for polygon: https://github.com/techmatters/terraso-backend/blob/67f6ee452d4bd2af51a64890da3a5b6f01eb27a4/terraso_backend/apps/core/models/landscapes.py#L30

terraso_backend/apps/graphql/schema/users.py
```diff
@@ -107,3 +118,38 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+        if not rules.test_rule("allowed_to_update_preferences", request_user, preference):
+            logger.error(
+                "Attempt to update a User preferences, not allowed",
+                extra={"request_user_id": request_user.id, "target_user_id": user.id},
+            )
+            raise GraphQLNotAllowedException(
+                model_name=UserPreference.__name__, operation=MutationTypes.UPDATE
+            )
+
+        preference, _ = UserPreference.objects.get_or_create(user_id=user.id, key=key)
```
cacarrara, line 151: Issue: the `UserPreference` is already been fetched on line 140.

terraso_backend/apps/core/models/users.py
cacarrara: Just for context, we have good conditions to make queries on `JSONField`: https://docs.djangoproject.com/en/4.0/topics/db/queries/#querying-jsonfield

josebui: @cacarraraThat is a good suggestion, as long as we are able to just query the specific preferences the UI needs, but I think that will require custom graphQL resolvers, right? or else it will be just one `preferences`field in the UserNode returning everything. So from the `models` side it will jsut be one extra field in the User model called preferences, but in the GraphQL schema it will have the same pieces, what do you think?

josebui: @cacarrara I do see some cases:
- Preferences related to lists pagination
- Preferences related to map size display
- Prefrences related to notifications
- Preferences related to emails
Those ideally should be fetched in the section or module where they are needed, but I do understand those are not requirements right now, but likely will be.

I think in general I have more concerns related to keeping the structure of the preferences field not controlled. For example: 
- Eventually we will want to generate emails from a different service and that service will require to know the locale selected by the user, since the api is not saying anything related to the structure then this will be a coordination between this new module or service and the web client related to the structure of the preferences field.
- Also, if a different service needs to set user preferences they will have a similar issue, they will have to coordinate at some level with the rest of the services to know what the complete structure is instead of just the key of the preference.

Another downside with the approach (if eventually we identify it is not enough) it will require data schema changes and data migration that are in general expensive and complex to coordinate, so a simplistic approach right now might produce a difficult to revert state.

So, I think we have some options:
1. We can keep preferences related to each service, like web client preferences, mailing service preferences, api preferences, etc.. but this will not let us reuse preference between services
2. Keep the current implementation and set the `value` field in `UserPreference` to be a `JSONField` this will give us more structure inside that value.
3. Change it to just one field in the User model but keep the structure from the GraphQL layer

What do you think?

cacarrara: Let's keep it as it is for now. :+1:

terraso_backend/apps/graphql/schema/__init__.py
```diff
@@ -30,7 +30,13 @@
     MembershipNode,
     MembershipUpdateMutation,
 )
-from .users import UserAddMutation, UserDeleteMutation, UserNode, UserUpdateMutation
+from .users import (
+    UserAddMutation,
+    UserDeleteMutation,
+    UserNode,
+    UserPreferenceUpdate,
```
paulschreiber, line 37: question: does this handle add, update and delete?

josebui: @paulschreiber not delete, but add and update it does, I can add a delete mutation too

terraso_backend/apps/graphql/schema/users.py
```diff
@@ -107,3 +123,44 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+                extra={"request_user_id": request_user.id, "target_user_id": user.id},
+            )
+            raise GraphQLNotAllowedException(
+                model_name=UserPreference.__name__, operation=MutationTypes.UPDATE
+            )
+
+        try:
+            value_json = json.loads(value_raw)
+        except json.JSONDecodeError:
+            raise GraphQLValidationException.from_custom_errors(
```
josebui, line 159: Hey @cacarrara How does this look to you? I changed the `value` field to a json field and I had to add this validation, there are some decisions that I am not completely sure are correct:
- The `bad_format`code is hardcoded here, should that exist in a constant file? or where are the other codes?
- I create a `from_custom_errors` function to handle this error, is that the correct way? or is there something else in the code to handle something like this?

terraso_backend/apps/core/migrations/0011_userpreference_userpreference_unique_user_preference.py
```diff
@@ -0,0 +1,48 @@
...
+                        related_name="preferences",
+                        to=settings.AUTH_USER_MODEL,
+                    ),
+                ),
+            ],
+        ),
+        migrations.AddConstraint(
+            model_name="userpreference",
+            constraint=models.UniqueConstraint(
+                condition=models.Q(("deleted_at__isnull", True)),
```
cacarrara, line 43: Friendly reminder: the migration need to be updated.

cacarrara (approved):
Check my comment on migration file

paulschreiber (approved):

PR #142, opened 2022-03-24
feat: add back-end session

cacarrara:
In order to integrate Django admin authentication with our current
authentication system, we have to use Django sessions. Also, in near
future we will have to use Django sessions to enable users authenticate
to other services using OAuth.

So, this change increments our current authentication process to
properly create Django sessions after user signs in. The only difference
in the sign in process is the call to Django's `login()`.

Also, to provide a mechanism to end sessions, this change adds a new
endpoint to the API `/logout`. This view simply call Django's `logout()`
function and returns `OK` to the client.


terraso_backend/apps/auth/views.py
```diff
@@ -139,12 +137,25 @@ def post(self, request, *args, **kwargs):
...
                 "access_token": access_token,
                 "refresh_token": refresh_token,
             }
         )
+
+
+class LogoutView(View):
+    def post(self, request, *args, **kwargs):
+        dj_logout(request)
+        return HttpResponse("OK", status=200)
```
josebui, line 153: Question: Will this be displayed to the user, or is the intent that a client application will call this and show the proper message to the user?

terraso_backend/tests/auth/test_views.py
```diff
@@ -46,6 +46,7 @@ def test_get_google_callback(client, access_tokens_google, respx_mock):
     assert response.status_code == 302
     assert response.cookies.get("atoken")
     assert response.cookies.get("rtoken")
+    assert response.cookies.get("sessionid")
```
josebui, line 49: Question: Is the session ID a replacement to the tokens? I mean how will the clients use this session ID?

terraso_backend/apps/auth/views.py
cacarrara: The intent is that a client application will call this and show the proper message to the user.

terraso_backend/tests/auth/test_views.py
cacarrara: The `sessionid` is not a replacement to the tokens. The web client won't be affected in any way by this new session. The `sessionid` token will only be used by the `/admin` views (and related views, like OAuth authorization).

It's important, though, that when the web client signs users out, it also calls to the `logout` endpoint to end the backend session. Otherwise, the user will sign out from the web client, but still be signed in to the admin pages. I created the issue so we can keep it tracked: https://github.com/techmatters/terraso-web-client/issues/281

Important: the admin pages will only be accessed by Terraso administrators, however the OAuth authorization page (that uses backend session) will be accessed by any user. When a non-staff (non-admin) user attempt to access the admin page, no resources are going to be showed.

paulschreiber (approved):

PR #156, opened 2022-04-28
feat: create the model layer of shared data app

cacarrara:
This change adds the model layer for the new `shared_data` app. The app
will be the central place for the "Shared Data Files" feature.


terraso_backend/apps/shared_data/models.py
```diff
@@ -0,0 +1,52 @@
...
+    only has the URL for that resource as a link to it.
+
+    Attributes
+    ----------
+    name: str
+        any user given name for that resource
+    description: str
+        a longer description explaining the resource
+    resource_type: str
+        the 'technical' type of the resource, usually the mime type
```
paulschreiber, line 25: question: do we have a map of MIME types to file extensions?

terraso_backend/apps/shared_data/permission_rules.py
```diff
@@ -0,0 +1,11 @@
+import rules
+
```
paulschreiber, line 2: question: Do we need an `allowed_to_view_entry` rule?

paulschreiber (approved):

terraso_backend/apps/shared_data/models.py
cacarrara: Python has, and it's going to be used when receiving the file: https://docs.python.org/3/library/mimetypes.html

PR #166, opened 2022-05-11
feat: add upload view to create data entries

cacarrara:
This change adds the API view to upload (and create) Shared Data Entries.


paulschreiber (changes requested):

terraso_backend/apps/shared_data/forms.py
```diff
@@ -0,0 +1,74 @@
...
+            and allowed_file_extensions
+            and file_extension not in allowed_file_extensions
+        ):
+            message = (
+                f"Invalid file extension ({file_extension}) for the file type {file_mime_type}"
+            )
+            logger.info(message)
+            raise ValidationError(message, code="invalid")
+
+        cleaned_data["resource_type"] = file_mime_type
```
josebui, line 60: Suggestion: I think the decision was to hold the extension in the `resource_type` for this first iteration, that way we can easily make decision in the client side on what to display depending on the extension, from our converastion last week it seems the mime_type is not as predictable as an extension, what do you think?

.env.sample
```diff
@@ -26,5 +26,11 @@ JWT_SECRET=superinsecuresecretchangeit
 
 PROFILE_IMAGES_S3_BUCKET=images.dev.terraso.org
 PROFILE_IMAGES_BASE_URL=https://images.dev.terraso.org
+
+DATA_ENTRY_FILE_S3_BUCKET=files.dev.terraso.org
+DATA_ENTRY_FILE_BASE_URL=https://files.dev.terraso.org
```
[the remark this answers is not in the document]
cacarrara: I followed the same pattern used on profile images above. I wanted to preserve consistency. We might review that later.

terraso_backend/config/settings.py
```diff
@@ -216,6 +216,10 @@
 
 PROFILE_IMAGES_S3_BUCKET = config("PROFILE_IMAGES_S3_BUCKET", default="")
 PROFILE_IMAGES_BASE_URL = config("PROFILE_IMAGES_BASE_URL", default="")
+
+DATA_ENTRY_FILE_S3_BUCKET = config("DATA_ENTRY_FILE_S3_BUCKET", default="")
+DATA_ENTRY_FILE_BASE_URL = config("DATA_ENTRY_FILE_BASE_URL", default="")
```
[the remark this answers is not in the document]
cacarrara: I followed the same pattern used on profile images above. I wanted to preserve consistency. We might review that later.

terraso_backend/apps/shared_data/forms.py
cacarrara: Thanks for the heads-up, Jose. It's been changed.

paulschreiber (approved):

terraso_backend/apps/graphql/schema/shared_data.py
```diff
@@ -39,6 +39,9 @@ class Meta:
         interfaces = (relay.Node,)
         connection_class = TerrasoConnection
 
+    def resolve_url(self, info):
+        return self.signed_url
```
josebui, line 43: Question: What is the difference between generating the S3 URL here or in the model?

josebui: I think if we add it here the code while only be called if the filed is in the query

paulschreiber: We don't want to store the signed URL in the database, since it's temporary.

josebui: @cacarrara Yes, I see that, my question is (since the signed URL is somwthing that is relevant to the clients) if it worth generating the signed URL at the GraphQL level to avoid generating it even if the url is not needed

PR #169, opened 2022-05-19
feat: restrict DataEntry list to group members only

cacarrara:
This change makes the GraphQL query for `DataEntry` return only those
registers associated to the Groups from the logged user.

### DIFF
--- a/terraso_backend/tests/graphql/conftest.py
+++ b/terraso_backend/tests/graphql/conftest.py
@@ -150,6 +150,26 @@ def make_core_db_records(
 
 
 @pytest.fixture
-def data_entries(users):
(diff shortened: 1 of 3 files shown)


terraso_backend/tests/graphql/conftest.py
```diff
@@ -150,6 +150,26 @@ def make_core_db_records(
 
 
 @pytest.fixture
-def data_entries(users):
+def data_entry_logged_user(users, groups):
```
paulschreiber, line 153:
```suggestion
def data_entry_current_user(users, groups):
```

paulschreiber (changes requested):

PR #174, opened 2022-05-25
feat: GraphQL error structure on upload Shared Files

caiocarrara:
This change uses the same error structure from GraphQL responses on
upload shared files response.


terraso_backend/apps/core/models/__init__.py
```diff
@@ -8,6 +9,7 @@
     "Landscape",
     "LandscapeGroup",
     "Membership",
+    "SlugModel",
```
josebui, line 12: Question: Shy is SlugModel here, does this means there a table for that in the DB? or why do we need it here?

cacarrara:
@josebui the branch was outdated... I updated it now.

josebui:
It looks good, I am going to hook it to the web client to see if we need more things from the backend

josebui (approved):

PR #180, opened 2022-06-02
feat: attempt to update profile image on every login

cacarrara:
This change updates the process of setting user's profile image to
attempt to download (and update) user's profile image on every login.

This is necessary to mitigate the following issue. So, user's profile
image can be updated without having to remove the account from 
Terraso:

- https://github.com/techmatters/terraso-web-client/issues/363

### DIFF
--- a/terraso_backend/apps/auth/services.py
+++ b/terraso_backend/apps/auth/services.py
@@ -46,10 +46,11 @@ class AccountService:
     def _persist_user(self, email, first_name="", last_name="", profile_image_url=None):
         user, created = User.objects.get_or_create(email=email)
 
+        self._update_profile_image(user, profile_image_url)
+
         if not created:
             return user
 
-        profile_image_service = ProfileImageService()
         update_name = first_name or last_name
 
         if first_name:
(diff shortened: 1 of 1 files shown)


terraso_backend/apps/auth/services.py
```diff
@@ -58,18 +59,24 @@ def _persist_user(self, email, first_name="", last_name="", profile_image_url=No
...
-        except Exception:
-            logger.exception("Failed to upload profile image. User ID: {}".format(user_id))
-
         if update_name:
             user.save()
 
         return user
 
+    def _update_profile_image(self, user, profile_image_url):
+        if not profile_image_url:
```
paulschreiber, line 68: question: won't this exit early with a missing/blank URL and prevent it from updating?

paulschreiber (changes requested):

terraso_backend/apps/auth/services.py
cacarrara: Yes, and it should. Since the following logins after the initial one the image is not provided by Google, if we update the URL with the empty value, we are going to erase the value on every login.

paulschreiber (approved):

PR #183, opened 2022-06-08
fix: Disallow "new" as a name for groups and landscapes

### LINKED ISSUE #164: Don't allow "new" as landscape or group slug
## Description
Don't allow  the use of "new" as landscape or group slug. This creates a route conflict with the app.

The landscape `/landscapes/new` would be inaccessible.

paulschreiber:
## Description
Disallow "new" as a name for groups and landscapes through a validator.

### Checklist
- [X] Corresponding issue has been opened
- [x] New tests added

### Related Issues
Fixes #164.

### DIFF
--- a/terraso_backend/tests/core/models/test_groups.py
+++ b/terraso_backend/tests/core/models/test_groups.py
@@ -1,5 +1,4 @@
 import pytest
-from django.conf import settings
 from django.core.exceptions import ValidationError
 from mixer.backend.django import mixer
 
@@ -47,7 +46,7 @@ def test_group_disallowed_name():
         group.full_clean()
 
 
-def test_group_additional_disallowed_name():
+def test_group_additional_disallowed_name(settings):
     group_name = "Foobar"
     settings.DISALLOWED_NAMES_LIST = ["new", "foobar"]
     group = mixer.blend(Group, name=group_name)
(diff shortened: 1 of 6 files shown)


cacarrara (commented):
It looks ok.

@paulschreiber, for testing it, you can create model test that creates new models with the name to be avoided and call the `full_clean()` method on the model's instance. More about model validation here: https://docs.djangoproject.com/en/4.0/ref/models/instances/#validating-objects

Also, you can override the `DISALLOWED_NAMES_LIST` during the test to make sure the process is using it as a source for words to be avoided. The pytest plugin we use has a fixture ready-to-use to change the settings: https://pytest-django.readthedocs.io/en/latest/helpers.html#settings

terraso_backend/tests/core/models/test_groups.py
```diff
@@ -20,6 +22,23 @@ def test_group_is_slugifiable_by_name():
...
 
+def test_group_disallowed_name():
+    group_name = "New"
+    group = mixer.blend(Group, name=group_name)
+
+    with pytest.raises(ValidationError, match="New is not allowed as a name"):
+        group.full_clean()
+
+
+def test_group_additional_disallowed_name():
```
cacarrara, line 33: The correct way to tweak settings in a test is receiving `settings` as fixture.

https://pytest-django.readthedocs.io/en/latest/helpers.html#settings

cacarrara (changes requested):

terraso_backend/tests/core/models/test_groups.py
paulschreiber: @cacarrara Can you make the change use the GitHub suggest feature?

cacarrara (approved):

PR #198, opened 2022-09-16
feat: enforce allowed file extensions list

### LINKED ISSUE #197: Restrict upload file types to only allowed types
## Description
Currently, the list of allowed file types/extensions is only enforced on the client. It should be enforced on the backend.

paulschreiber:
## Description
Restrict file uploads to allowed extensions list.

### Related Issues
Fixes #197.


paulschreiber:
@josebui it would be good to use separate lists for uploading map data (json/geojson) and shared files (xls/pdf/doc/ppt). Do you know how we would do that?

josebui:
@paulschreiber What do you mean by different lists? you mean in terms of splitting the list `DATA_ENTRY_ACCEPTED_EXTENSIONS` list? or having different endpoints to handle the different groups? right now there is still no need to store map data as files, my recommendation will be to handle that split of logic when that requirement appears

josebui:
@paulschreiber Question: Do we have tests already in the backend to test the extensions validation?

paulschreiber:
@josebui when you upload files for shared files, we support Word/Excel/CSV/PPT/PDF. When you upload files for map boundaries, we support .json/.geojson. Right now, everything is in `DATA_ENTRY_ACCEPTED_EXTENSIONS` and it would be good to have two lists so the server prevents uploads of .json files as shared files.

josebui:
@paulschreiber Ohh, if that is the case, and we are using the same list for boundaries we should defenitely split it, we should have different configuration varaibles for now. Boundaries are not files that we upload to S3

josebui:
@paulschreiber This PR is not functionaly ready, I don't think we need the split right now since the boundaries are always stored as GeoJSON so we just need a list for the extensions accepted by the upload feature, I think this can be handled as part of https://trello.com/c/CWQsLJm4/52-upload-map-files-to-files-module

PR #205, opened 2022-10-03
feat: track user signup on backend

david-code:
## Description

The frontend has to track signups when they occur ([signup tracking issue](https://github.com/techmatters/terraso-web-client/issues/492)). Unfortunately this is not really possible right now because the OAuth callback is handled on the backend, so the frontend doesn't really know if a new user has been created after the user clicks on the signup button. The idea here is that we can set the cookie when a new user is created, the cookie can be read on the frontend, a request will be sent to plausible to track which service was used to create the user, and finally the cookie will be discarded.

Let me know if this seems OK to you, or if you think there is a better way.

### Checklist
- [ ] Should an issue be added?


josebui:
@david-code Question: Is there a way to send events to plausible from the backend? I think we want to avoid as much as we can cookies that track behavior attached to specific users in the client side

david-code:
> Question: Is there a way to send events to plausible from the backend (I think they have an API)? I think we want to avoid as much as we can cookies that track behavior attached to specific users in the client side

I looked briefly into this and I didn't find a python client for the API, so I gave up. However I just looked up the [Plausible Event API](https://plausible.io/docs/events-api) and it looks very simple to use. I agree, this would probably be a better solution. I'll try and use the API instead and see how it goes.

terraso_backend/apps/auth/services.py
```diff
@@ -108,3 +112,89 @@ def _get_base_payload(self, user):
...
+        # this is a mandatory header for the Plausible API, see docs for details
+        ip_address = "127.0.0.1"
+        if "x-forwarded-for" in req.headers:
+            ip_address = self._get_first_ip_address(req.headers["x-forwarded-for"])
+            if not ip_address:
+                logger.error(
+                    "During signup tracking, request header 'x-forwarded-for' was set,"
+                    " but no valid ip addresses were found"
+                )
+                return
```
david-code, line 195: Need to send the `x-forwarded-for` header for the [Plausible API](https://plausible.io/docs/events-api). Will need to check if this header is being received on the deployments, or else this won't quite work as expected.

paulschreiber:
Would be good to file a bug report on the library about the lack of a retry limit.

terraso_backend/apps/auth/views.py
```diff
@@ -72,19 +74,51 @@ def process_callback(self):
...
+            try:
+                plausible_service.track_signup(created_with_service, self.request)
+            except httpx.HTTPStatusError as e:
+                logger.error(
+                    "Error tracking signup: received status code %s when querying %s",
+                    e.response.status_code,
+                    e.request.url,
+                )
+            except Exception:
+                logger.exception("Error tracking signup")
```
paulschreiber, line 89:
```suggestion
            except Exception as e:
                logger.exception("Error tracking signup: " +str(e))
```

paulschreiber: question: would it make sense to include details about the exception

paulschreiber (changes requested):

terraso_backend/apps/auth/services.py
```diff
@@ -108,3 +112,89 @@ def _get_base_payload(self, user):
...
         }
+
+
+class PlausibleService:
+    """Service for making API calls to plausible service.
+
+    See documentation at https://plausible.io/docs/events-api .
+    """
+
+    PLAUSIBLE_URL = "https://plausible.io/api/event"
```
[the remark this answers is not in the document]
david-code: This is the only instance of the API, and there's no way that we could run a local instance so the value isn't going to change. However it could make sense to keep these kinds of values outside of the code in general, it is a bit neater. I think I will move the value to the settings, unless you have some objections.

terraso_backend/apps/auth/views.py
david-code: Apparently the Python `logger.exception` does this already ([documentation](https://docs.python.org/3/library/logging.html#logging.Logger.exception)). I learned this doing this change actually.

terraso_backend/apps/auth/services.py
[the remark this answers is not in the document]
paulschreiber: Plausible is open source and can be self-hosted. https://github.com/plausible/analytics

.env.sample
```diff
@@ -34,3 +34,4 @@ AWS_ACCESS_KEY_ID=aws-key-id
 AWS_SECRET_ACCESS_KEY=aws-secret-access-key
 
 CDN_STATIC_DOMAIN=static.terraso.org
+PLAUSIBLE_URL=https://plausible.io/api/event
```
paulschreiber, line 37: issue: Please add a newline to the end of the file.

paulschreiber, line 37:
```suggestion
PLAUSIBLE_URL=https://plausible.io/api/event

```

paulschreiber (approved):

PR #212, opened 2022-10-24
feat: Add option for reset/backup from different environment to Django admin

david-code:
## Description

This PR adds a reset button to the Django admin interface that can be optionally enabled. When pressed, it will trigger the following actions:

- S3 buckets from the source environment (A) will be synced with buckets in the target environment (B)
- A job will be triggered on A's Backend Service to create a backup archive and upload it to another S3 bucket.
- This archive will be downloaded by B's Backend Service. B's database will be wiped and restored from the archive. For all object in B's Database, URLs pointing to S3 buckets in A will be updated to point to the corresponding URLs in B.

There are a few tricky parts here.

- We use the database to save the progress of the long-running jobs, but the database needs to be wiped. This means that the background jobs table cannot be wiped.
- While we delete the database, the user will continue to query back to an endpoint. This means that Django's authentication will not work as we want it to if the user queries before the user object is re-added to the database.


paulschreiber:
How do you want to handle cases where the hostname changes between instances? That will affect stored URLs.

david-code:
> How do you want to handle cases where the hostname changes between instances? That will affect stored URLs.

Good point, I hadn't thought of that. I will take another look and see what the options could be.

terraso_backend/apps/core/management/commands/backup.py
```diff
@@ -0,0 +1,83 @@
...
+                    storage.save(data_file_name, data_fp)
+                    storage.save(migration_file_name, migration_fp)
+
+        except Exception:
+            traceback.print_exc()
+        finally:
+            if tempdir:
+                tempdir.cleanup()
+            if options["s3"]:
+                # if we are uploading the backups, we don't want extra files
```
paulschreiber, line 82: issue: please explain what you are cleaning up.

terraso_backend/apps/core/management/commands/backup.py
```diff
@@ -0,0 +1,83 @@
...
+                """SELECT DISTINCT ON (app) app, name
+                   FROM django_migrations
+                    ORDER BY app, name DESC"""
+            )
+            rows = cursor.fetchall()
+        return dict(rows)
+
+    @staticmethod
+    def _save_migration_versions(filename, migration_versions):
+        """Write the migration versions to a JSON file. For now."""
```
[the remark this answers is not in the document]
david-code: I think I originally wanted to write the migrations JSON directly to the backup JSON file. I realized later that this would be way more trouble than it's worth. I'll remove the comment.

terraso_backend/apps/core/management/commands/loadbackup.py
```diff
@@ -0,0 +1,101 @@
...
+        # default is to sort ascending, so latest files at end
+        return files[-2:]
+
+    @staticmethod
+    def _copy_s3_file(storage, path, suffix):
+        """Copies the S3 file to a temporary file on local filesystem."""
+        _, tmp_file_path = mkstemp(suffix=suffix)
+        with storage.open(path, "rb") as fp, open(tmp_file_path, "wb") as temp_file:
+            for chunk in fp.chunks():
+                temp_file.write(chunk)
```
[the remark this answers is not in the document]
david-code: I used the [django-storages](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html) library here because I saw it used in other parts of the code. This implements the Django storage API, which doesn't provide an API to download files to the local file system. However, poking around a bit, there is a way to use the underlying `boto3` library, which should allow this. I will look into this more.

terraso_backend/apps/core/management/commands/loadbackup.py
```diff
@@ -0,0 +1,101 @@
...
+    def _find_latest_backup_dir(directory):
+        """Find the latest backup in a directory.
+
+        Backups are named according to the date they are created.
+        """
+        files = sorted([f for f in directory.glob("backup*") if f.is_file()])
+        if len(files) < 2:
+            raise RuntimeError(f"Couldn't find any backup files in {directory}")
+        # default is to sort ascending, so latest files at end
+        return files[-2:]
```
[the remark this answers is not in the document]
david-code: It's just to grab the migrations JSON and backup JSON file. I will make this more explicit (I think it is in the other method)

terraso_backend/apps/core/management/commands/loadbackup.py
```diff
@@ -39,7 +39,8 @@ def _find_latest_backup_dir(directory):
         if len(files) < 2:
             raise RuntimeError(f"Couldn't find any backup files in {directory}")
         # default is to sort ascending, so latest files at end
-        return files[-2:]
+        data_file, migrations_file = files[-2:]
+        return data_file, migrations_file
```
paulschreiber, line 43: question: what else is in the files array?

david-code: It's possible that we have more than one set of backups in the bucket, for convenience or if there is a backup we want to save for whatever reason. Do you think there is a better way to handle this?

.env.sample
```diff
@@ -30,6 +30,8 @@ PROFILE_IMAGES_BASE_URL=https://images.dev.terraso.org
 DATA_ENTRY_FILE_S3_BUCKET=files.dev.terraso.org
 DATA_ENTRY_FILE_BASE_URL=https://files.dev.terraso.org
 
+DB_BACKUP_S3_BUCKET=backup-bucket
```
josebui, line 33: Suggestion: Can we have a `dev` bucket name named like the others, the have this structure: `files.dev.terraso.org`

josebui:
Question: @david-code Can you add a comment with some example usages of the 2 commands? or maybe even better can you add those to the README?

josebui:
Question: @david-code Can this hold multiple backups for one environment in the same S3 bucket?

david-code:
> Question: @david-code Can you add a comment with some example usages of the 2 commands? or maybe even better can you add those to the README?

Yes, good idea. I will do that.

> Question: @david-code Can this hold multiple backups for one environment in the same S3 bucket?

Yes, but right now it will only use the most recent files. The files are named according to the date created and can be sorted alphabetically. I was thinking of adding an option to the backup script to allow deleting olders backups so we don't have to do that by hand.

david-code:
> How do you want to handle cases where the hostname changes between instances? That will affect stored URLs.

I have added a new commit to address the issue of rewriting the URLs in the database. It is a bit complicated, and will require basically copying the S3 buckets to the restore instance buckets afterwards. I'll add some information in the README.

backup/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM debian:buster-slim
+
+RUN apt-get update && apt-get install -y curl unzip jq \
+          && curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" \
+          && unzip awscliv2.zip \
+          && ./aws/install
+
+COPY backup.sh /script/backup.sh
+
+ENTRYPOINT ["/script/backup.sh"]
```
paulschreiber, line 10:
```suggestion
ENTRYPOINT ["/script/backup.sh"]

```

backup/url_rewrites.conf
```diff
@@ -0,0 +1,7 @@
+[images]
+name = PROFILE_IMAGES_BASE_URL
+url_regex = ^images.terraso.org$
+
+[datasets]
+name = DATA_ENTRY_FILE_BASE_URL
+url_regex = ^files.terraso.org$
```
paulschreiber, line 7: issue: add newline to end of file

backup/env.sample
```diff
@@ -0,0 +1,6 @@
+RENDER_TOKEN=
+URL_REWRITES_CONFIG_FILE=/etc/terraso/url-rewrites.conf
+SOURCE_S3_BUCKETS=files.terraso.org,images.terraso.org
+TARGET_S3_BUCKETS=files.testing.terraso.net,images.testing.terraso.net
+SOURCE_SERVICE=
+TARGET_SERVICE=
```
paulschreiber, line 6: issue: add newline to end of file

terraso_backend/config/settings.py
```diff
@@ -223,6 +223,8 @@
 DATA_ENTRY_FILE_S3_BUCKET = config("DATA_ENTRY_FILE_S3_BUCKET", default="")
 DATA_ENTRY_FILE_BASE_URL = config("DATA_ENTRY_FILE_BASE_URL", default="")
 
+BACKUP_S3_BUCKET = config("BACKUP_S3_BUCKET", default="")
```
paulschreiber, line 226: issue: update to DB_BACKUP_S3_BUCKET

paulschreiber (changes requested):

terraso_backend/custom_templates/admin/index.html
```diff
@@ -0,0 +1,68 @@
...
+      }).then(() => {
+        statusP.textContent = "Job finished successfully.";
+      }).finally(() => {
+        if (intervalId) {
+          clearInterval(intervalId);
+        }
+      });
+
+    });
+  })();
```
paulschreiber, line 66: Please put this in a standalone JS file so it can be linted/minified.

terraso_backend/custom_templates/admin/index.html
```diff
@@ -0,0 +1,68 @@
...
+  {% endif %}
+</div>
+<script>
+  (function () {
+
+    const button = document.getElementById("restore-button");
+    const statusP = document.getElementById("restore-status");
+
+    const headers = {
+      "Cookie": document.cookie,
```
paulschreiber, line 18: question: do you need to add this manually?

terraso_backend/apps/core/views.py
```diff
@@ -32,14 +37,52 @@ def check_db_access():
...
+            json.dumps({"message": f"No task with id {task_id}"}), "application/json"
+        )
+    return HttpResponse(json.dumps({"status": task.status}), "application/json")
+
 
-    status = "running" if random.random() > 0.4 else "done"
-    return HttpResponse(json.dumps({"status": status}), "application/json")
+def restore(task):
+    try:
+        time.sleep(15)
```
paulschreiber, line 80: question: where does 15 seconds come from?
suggestion: make `SLEEP_TIME` a constant.

terraso_backend/apps/core/views.py
```diff
@@ -32,14 +37,52 @@ def check_db_access():
...
+    except Exception:
+        logger.exception("Job failed!")
+        print("FAILURE")
```
paulschreiber, line 83:
```suggestion
```

paulschreiber (changes requested):

terraso_backend/apps/core/static/instance_restore.js
```diff
@@ -0,0 +1,74 @@
+(function () {
+
+  const button = document.getElementById("restore-button");
+  const statusP = document.getElementById("restore-status");
+
+  function getCookie(name) {
+    let cookieValue = null;
+    if (document.cookie && document.cookie !== '') {
```
paulschreiber, line 8: Won't `if (document.cookie)` do the same thing?

terraso_backend/apps/core/static/instance_restore.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+
+  button.addEventListener("click", async () => {
+    const resp = await fetch("/admin/restore", { method: "POST", headers });
+    if (!resp.ok) {
+      setError();
+    }
+    const { taskId } = await resp.json();
+    statusP.textContent = "Job triggered";
```
paulschreiber, line 56: suggestion: 'Job started' (or other nontechnical language).

terraso_backend/apps/core/management/commands/loadbackup.py
```diff
@@ -0,0 +1,275 @@
...
+                    user.email = str(user.id)
+                    user.save()
+                    self._reset_user_id(new_user.id, user.id)
+                    # copy fields from new user to old
+                    for field in User._meta.fields:
+                        if field.name in ("id", "password", "is_staff"):
+                            continue
+                        setattr(user, field.name, getattr(new_user, field.name))
+                    new_user.delete()
+                    user.save()
```
david-code, line 253: This code creates a new user with the same ID as the currently logged in user. This is just to make sure that the user ID of the user triggering the reset does not change and they are not logged out. We copy any other information that might have changed from the user in the DB being restored.

terraso_backend/apps/core/management/commands/loadbackup.py
```diff
@@ -0,0 +1,275 @@
...
+                msg = "Command failed resetting database"
+                logger.exception(msg)
+                connection.rollback()
+                cleanup()
+                raise CommandError(msg)
+
+            connection.commit()
+
+        try:
+            core_skip = [23, 25]
```
david-code, line 211: These migrations need to be skipped. `23` adds the background tasks table, but this table needs to not be deleted. `25` adds taxonomy terms that are already present in the data restore. In the future, we can add other migrations that need to be skipped here.

terraso_backend/apps/core/views.py
```diff
@@ -23,3 +40,131 @@ def check_db_access():
...
+        session_user_id = session.get_decoded().get("_auth_user_id")
+        if session_user_id != str(task.created_by_id):
+            return HttpResponseNotAllowed(
+                json.dumps({"message": "You did not create this task"}), "application/json"
+            )
+    except BackgroundTask.DoesNotExist:
+        return HttpResponseNotFound(
+            json.dumps({"message": f"No task with id {task_id}"}), "application/json"
+        )
+    return HttpResponse(json.dumps({"status": task.status}), "application/json")
```
david-code, line 90: The user triggering the restore job will be deleted from the database at some point. This endpoint checks the cookie to see if the user is the person who triggered the job, and only returns the status if this is the case.

.env.sample
```diff
@@ -28,8 +28,12 @@ PROFILE_IMAGES_S3_BUCKET=images.dev.terraso.org
...
 
+DB_BACKUP_S3_BUCKET=backup.dev.terraso.org
+ALLOW_RESTORE_FROM_BACKUP=false
+DB_RESTORE_CONFIG_FILE=/etc/terraso/restore.conf
 AWS_ACCESS_KEY_ID=aws-key-id
 AWS_SECRET_ACCESS_KEY=aws-secret-access-key
 
 CDN_STATIC_DOMAIN=static.terraso.org
 PLAUSIBLE_URL=https://plausible.io/api/event
+RENDER_API_TOKEN=your-render-token
```
josebui, line 39: Issue: Missing the last line break

terraso_backend/apps/core/management/README.md
```diff
@@ -0,0 +1,34 @@
...
+### Source Instance
+
+- `DB_BACKUP_S3_BUCKET` must point to a S3 bucket where the backups will be stored
+- The `AWS_*` credentials must be valid
+
+
+### Target Instance
+
+- The `ALLOW_RESTORE_FROM_BACKUP` should be set to `true`. This will make the button to trigger the job visible in the Django admin panel.
+- The `DB_RESTORE_CONFIG_FILE` should point to the local of the restore configuration file.
```
josebui, line 24: Question: Why is the DB_RESTORE_CONFIG_FILE value configurable? In what situations will this be changed?

Dockerfile
```diff
@@ -5,7 +5,10 @@ RUN adduser --disabled-password terraso
 ENV PATH /home/terraso/.local/bin:$PATH
 
 RUN apt-get update && \
-    apt-get install -q -y --no-install-recommends build-essential libpq-dev dnsutils libmagic-dev mailcap && \
+    apt-get install -q -y --no-install-recommends build-essential libpq-dev dnsutils libmagic-dev mailcap unzip curl && \
+    curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
```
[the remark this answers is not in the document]
david-code: Good point, I will check it out.

terraso_backend/apps/core/models/landscapes.py
```diff
@@ -40,6 +43,26 @@ class Landscape(SlugModel):
     )
     groups = models.ManyToManyField(Group, through="LandscapeGroup")
 
+    area_types = models.JSONField(blank=True, null=True)
```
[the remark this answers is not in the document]
david-code: I think I rebased onto staging to make sure that I had the most recent changes for testing. This is a good point. It might make sense to merge this after the profile landscape changes are merged to main.

terraso_backend/apps/core/management/README.md
david-code: My reasoning was
- Don't want to hard code URLs into the code
- Don't want to add like 5 new env variables to the config file just for the backup, which will only be used on the testing environment

It is pretty easy to deploy it and is just a bit cleaner in my opinion.

david-code:
I got rid of the earlier commits that I got when I rebase on the `staging` branch. However this means that the migration I introduced will probably cause problems when Jose's PR for the landscape profile is merged. @josebui just a heads up on this.

terraso_backend/custom_templates/admin/index.html
```diff
@@ -0,0 +1,13 @@
+{% extends "admin/index.html" %}
+{% load static %}
+{% block content %}
+<div id="content-main">
+  {% include "admin/app_list.html" with app_list=app_list show_changelinks=True %}
+  {% if allow_restore %}
+    <button id="restore-button" type="button">Reload data</button>
```
paulschreiber, line 7:
```suggestion
    <button id="restore-button" type="button">Load data</button>
```

terraso_backend/apps/core/management/commands/loadbackup.py
```diff
@@ -0,0 +1,279 @@
...
+                    "sessions.Session",
+                    "admin.LogEntry",
+                ],
+            )
+
+            if user:
+                try:
+                    new_user = User.objects.get(email=user.email)
+                    # provide a fake "email" not in the database
+                    user.email = str(user.id)
```
paulschreiber, line 248: issue: please write a detailed comment explaining what is going on with the user ids, which tables are skipped, etc.

terraso_backend/apps/core/views.py
```diff
@@ -23,3 +40,130 @@ def check_db_access():
...
+        for key, value in zip(
+            ["AWS_ACCESS_KEY", "AWS_SECRET_ACCESS_KEY", "AWS_DEFAULT_REGION"], aws_credentials
+        )
+    }
+    env.update(os.environ)
+    for source, dest in zip(source_buckets, dest_buckets):
+        subprocess.run(["aws", "s3", "sync", "s3://" + source, "s3://" + dest], env=env)
+
+
+JOB_WAIT_TIME_SEC = 5
```
paulschreiber, line 106: suggestion: move this to the top of the file and give it a more descriptive name.

terraso_backend/apps/core/views.py
```diff
@@ -23,3 +40,130 @@ def check_db_access():
...
+            settings.AWS_SECRET_ACCESS_KEY,
+            settings.AWS_S3_REGION_NAME,
+        )
+        _sync_s3_buckets(source_buckets, dest_buckets, aws_credentials)
+        _backup_service(
+            service_name, settings.RENDER_API_TOKEN, "python3 terraso_backend/manage.py backup --s3"
+        )
+        _restore_from_backup(user_id, session_id)
+    except Exception:
+        logger.exception("Job failed!")
```
paulschreiber, line 164: suggestion: if you are going to leave this in production, add more details.

paulschreiber (changes requested):

terraso_backend/apps/core/static/instance_restore.js
david-code: I think so, but I copied this function directly from the [Django docs](https://docs.djangoproject.com/en/4.1/howto/csrf/#acquiring-the-token-if-csrf-use-sessions-and-csrf-cookie-httponly-are-false) so I'm not sure if there's a corner case that the double check covers.

terraso_backend/custom_templates/admin/index.html
```diff
@@ -0,0 +1,13 @@
+{% extends "admin/index.html" %}
+{% load static %}
+{% block content %}
+<div id="content-main">
+  {% include "admin/app_list.html" with app_list=app_list show_changelinks=True %}
+  {% if allow_restore %}
+    <button id="restore-button" type="button">Load data</button>
+    <p>Replace data data from another Terraso instance.</p>
```
paulschreiber, line 8:
```suggestion
    <p>Replace with data from another Terraso instance.</p>
```

terraso_backend/apps/core/management/commands/loadbackup.py
```diff
@@ -0,0 +1,282 @@
...
+                    "auth.Permission",
+                    "sessions.Session",
+                    "admin.LogEntry",
+                ],
+            )
+
+            if user:
+                try:
+                    # here we want to ensure that the currently logged in user (i.e. the user
+                    # triggering the reset) retains the same user ID in the restored data base.
```
paulschreiber, line 247:
```suggestion
                    # triggering the reset) retains the same user ID in the restored database.
```

terraso_backend/apps/core/management/commands/loadbackup.py
```diff
@@ -0,0 +1,282 @@
...
+                session.save()
+
+            cleanup()
+
+            config = Path(settings.DB_RESTORE_CONFIG_FILE)
+            if not config.is_file() or not os.access(config, os.R_OK):
+                raise CommandError(
+                    f"Path supplied for URL rewrites is not a file: {str(config)}. "
+                    "URL rewrites not applied."
+                )
```
paulschreiber, line 275:
```suggestion
            if not config.is_file() :
                raise CommandError(
                    f"Path supplied for URL rewrites is not a file: {str(config)}. "
                    "URL rewrites not applied."
                )
            if not os.access(config, os.R_OK) :
                raise CommandError(
                    f"Cannot read config file: {str(config)}. "
                    "URL rewrites not applied."
                )

```

paulschreiber (changes requested):

paulschreiber (approved):

PR #223, opened 2022-11-24
feat: Add changes to delete landscape profile photo

david-code:
Just a small change that lets us change the profile photo values from the frontend.


josebui (approved):
Looks good.
Suggestion: maybe just a test to validate the profile image can be deleted

david-code:
Added a test to
- add profile image
- delete profile image
- make sure updating other fields doesn't reset profile images

PR #224, opened 2022-11-28
feat: Share links

josebui:
## Description

- Added entry type field
- Added mutation to create shared data entry
- Added condition to create signe URL for S3

### Checklist
- [x] New tests added


terraso_backend/apps/graphql/schema/data_entries.py
```diff
@@ -48,7 +50,46 @@ def get_queryset(cls, queryset, info):
...
+            kwargs["created_by"] = user
+
+        if "entry_type" in kwargs:
+            kwargs["entry_type"] = DataEntry.get_entry_type_from_text(kwargs["entry_type"])
+
+        group_slug = kwargs.pop("group_slug")
+        try:
+            group = Group.objects.get(slug=group_slug)
+        except Group.DoesNotExist:
+            logger.error("Group not found when adding Data Entry", extra={"group_slug": group_slug})
```
paulschreiber, line 85: question: do we want to use `logger.exception`?

david-code:
Looks good to me!

terraso_backend/apps/graphql/schema/data_entries.py
josebui: @paulschreiber I don't think so, it seems that that log will give us all the information that we need, I don't think we need the stack trace, or what do you think? should I change it?

paulschreiber: @david-code what do you think?

david-code: I think it's fine just to log the message here. Because the exception caught is `Group.DoesNotExist`, the stack trace will just let us know that the group doesn't exist and shouldn't have any extra information.

paulschreiber (approved):

david-code (approved):

PR #231, opened 2022-12-19
feat: Extend boundary file formats (KML, KMZ, Shapefile) 

josebui:
## Description

- Process kml, kmz and shapefiles and transform them to geojson

### Checklist
- [x] New tests added

### Related Issues
Fixes #https://github.com/techmatters/terraso-web-client/issues/641


terraso_backend/apps/core/gis/parsers.py
```diff
@@ -0,0 +1,47 @@
...
+import os
+import uuid
+import zipfile
+
+import geopandas as gpd
+
+from apps.core.gis.utils import DEFAULT_CRS
+
+
+def isShapefile(file):
```
paulschreiber, line 11: suggestion: rename this function so the intent is more clear, i.e. you are only checking the extension, not the contents. perhaps hasShapefileExtension or isPossibleShapefile.

terraso_backend/apps/core/urls.py
```diff
@@ -1,11 +1,18 @@
...
+    create_restore_job,
+)
 
 app_name = "apps.core"
 
 urlpatterns = [
     path("healthz/", HealthView.as_view(), name="healthz"),
     path("admin/restore", create_restore_job),
     path("admin/restore/jobs/<int:task_id>", check_restore_job_status),
+    path("gis/parseFile/", csrf_exempt(ParseGeoFileView.as_view()), name="parse"),
```
paulschreiber, line 17:
```suggestion
    path("gis/parse/", csrf_exempt(ParseGeoFileView.as_view()), name="parse"),
```

terraso_backend/apps/core/views.py
```diff
@@ -14,16 +14,55 @@
...
+class ParseGeoFileView(AuthenticationRequiredMixin, FormView):
+    def post(self, request, **kwargs):
+        file = request.FILES.get("file")
+
+        geojson = None
+        if isShapefile(file):
+            try:
+                geojson = parseShapefile(file)
+            except Exception as e:
+                logger.exception("Error when parsing shapefile", error=e)
```
paulschreiber, line 46: question: does the log include the file name?

terraso_backend/config/settings.py
```diff
@@ -44,6 +44,7 @@
     "apps.graphql",
     "apps.auth",
     "apps.shared_data",
+    "geopandas",
```
david-code, line 47: questions: Why is necessary to add `"geopandas"` here, is there some kind of integration with Django?

david-code (commented):
Looks good!

I have a question about the format of the Shapefile zip: how did you decide on requiring the three file extensions `.shp`, `.shx` and '.prj'? Was this something that the users expect? I ask because I checked out the Wikipedia page for shapefiles and it seemed that `.prj` is not listed as a "required" file. I was wondering if would make sense to handle the case where the user does not upload this file, or if we are expecting them to always include the file from some examples or past agreement.

terraso_backend/config/settings.py
josebui: @david-code I followed a tutorial online, that is why I added it here, I will check if I can remove it

josebui:
> Looks good!
> 
> I have a question about the format of the Shapefile zip: how did you decide on requiring the three file extensions `.shp`, `.shx` and '.prj'? Was this something that the users expect? I ask because I checked out the Wikipedia page for shapefiles and it seemed that `.prj` is not listed as a "required" file. I was wondering if would make sense to handle the case where the user does not upload this file, or if we are expecting them to always include the file from some examples or past agreement.

@david-code My assumption here, from the files that Derek sent me is that the projection definition is critical. It is not required for the format (as far as I understand) because the shapefile doesn't need to necessarily be renderend on top of a map, in our case it is required for the system to render the data on top of a map, let me know if this makes sense

david-code:
> @david-code My assumption here, from the files that Derek sent me is that the projection definition is critical. It is not required for the format (as far as I understand) because the shapefile doesn't need to necessarily be renderend on top of a map, in our case it is required for the system to render the data on top of a map, let me know if this makes sense

Yep, that makes perfect sense.

terraso_backend/apps/core/views.py
josebui: @paulschreiber No, but I just added it

terraso_backend/apps/core/views.py
```diff
@@ -14,16 +14,58 @@
...
+                logger.exception(f"Error when parsing shapefile. File name: {file.name}", error=e)
+                return JsonResponse(
+                    {"errors": [{"message": json.dumps([{"code": "invalid_shapefile"}])}]},
+                    status=400,
+                )
+        elif is_kml_file_extension(file):
+            try:
+                geojson = parse_kml_file(file)
+            except Exception as e:
+                logger.exception("Error when parsing KML file", error=e)
```
paulschreiber, line 58:
```suggestion
                logger.exception("Error when parsing KML file. File name: {file.name}", error=e)
```

paulschreiber (approved):
Approved. Please accept (or modify) suggestions before merging.

david-code (approved):

PR #237, opened 2023-01-09
fix: Fix error message for duplicate landscape name creation

### LINKED ISSUE #233: bug: Validation error message for duplicated landscape name does not indicate which fields are failing
## Description

When the user tries to create a landscape with a name that has already been taken, a generic error message is displayed.

image

This was introduced by the changes in `aa7e57e`. Because the unique constraint is defined at the table level, the validation is no longer able to determine which column has triggered the unique constraint. As a result the error message is too vague.

## Steps To Reproduce
1. Try creating a landscape with the same name as an already existing landscape

## Expected behavior

An error message will indicate that the landscape name already exists

## Actual behavior

A generic error message is displayed.

## Additional context

[django-safe-delete docs](https://django-safedelete.readthedocs.io/en/latest/models.html#fields-uniqueness)

david-code:
## Description

This PR modifies the base model classes extended to easily generate table-level unique constraints for the model, while still providing descriptive error messages.

Any model extending BaseModel and BaseModelMeta can declare a list of `_unique_fields`. A custom metaclass uses these fields to define unique table constraints that respect what is needed by the `django-safe-delete` plugin.

### Checklist
- [x] New tests added

### Related Issues
Fixes #233 

### Verification steps

- Try creating a landscape with a name that already exists
- The error message should say "Landscape name must be unique."


paulschreiber (commented):
Will defer to @josebui.

terraso_backend/apps/core/models/taxonomy_terms.py
```diff
@@ -41,6 +41,7 @@ def __str__(self):
         return "{}.{}".format(self.type, self.slug)
 
     class Meta(SlugModel.Meta):
+        _ignore_unique_fields = ["slug"]
```
josebui, line 44: Question: I am not sure if I am able to understand all the changes, maybe this question will help. What is the difference between having here: `_unique_fields = []` or `_ignore_unique_fields = ["slug"]`?

terraso_backend/apps/core/migrations/0036_remove_group_core_group_unique_active_slug_and_more.py
```diff
@@ -0,0 +1,119 @@
...
+import apps.core.models.commons
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("core", "0035_landscapedevelopmentstrategy_opportunities"),
+    ]
+
+    operations = [
```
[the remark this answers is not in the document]
david-code: Oh, OK. I think I will also make a bug to look at why the `deleted_at` changes always appear.

terraso_backend/apps/core/models/taxonomy_terms.py
david-code: Yes, that's a good question. 

The metaclass takes the unique fields from each parent class to gather the list of constraints it should create. 

```python
        all_unique_fields = set(
            attrs.get("_unique_fields", [])
            + [field for base in bases for field in getattr(base, "_unique_fields", [])]
        )
```

I did this so that we could concatenate the lists throughout the class category, rather than the child class overwriting the parent class. However that means that there is no obvious way for the child class to skip over a unique field listed in the parent class. An easy solution is just to list explicitly if you would like to ignore a unique field declared in the parent class.

josebui (approved):

terraso_backend/apps/core/models/taxonomy_terms.py
josebui: Thanks David

PR #238, opened 2023-01-09
fix: delete users with shared files

### LINKED ISSUE #190: Cannot delete users who have uploaded shared files
## Description
Attempting to delete a user who has uploaded files will give an error like so

```
Cannot delete user
Deleting the selected user would require deleting the following protected related objects:

Data entry: acBie9x4 WieezMsPbKL4P2
Data entry: KoBo question set
Data entry: myfile
Data entry: plus+sign+cool
Data entry: acBie9x4WieezMsPbKL4P2

```

david-code:
## Description

Allows a user to be deleted, even if they have associated data files.

Follows the following logic:

- On a soft delete: linked `data entry` not affected
- On a hard delete: `created_by` field in linked data entry set to null.

Here is the [reference on soft delete policies](https://django-safedelete.readthedocs.io/en/latest/models.html#policies) that will be helpful in code review.

### Checklist
- [x] New tests added

### Related Issues
Fixes #190

### Verification steps

- Try deleting a user with associated data files

### DIFF
--- a/terraso_backend/apps/core/models/users.py
+++ b/terraso_backend/apps/core/models/users.py
@@ -85,6 +85,17 @@ class User(SafeDeleteModel, AbstractUser):
             .exists()
         )
 
+    def soft_delete_policy_action(self, **kwargs):
+        """Relink files to deleted user. The default policy is to set the `created_by` field to
+        null if the user is deleted. However, for a soft deletion we want to keep this link. That
+        way if the user is restored, the created_by is still pointing to the same place."""
+        linked_dataentries = self.dataentry_set.all()
+        delete_response = super().soft_delete_policy_action()
+        for entry in linked_dataentries:
+            entry.created_by = self
+            entry.save()
+        return delete_response
+
     def is_group_manager(self, group_id):
         return self.memberships.managers_only().filter(group__pk=group_id).exists()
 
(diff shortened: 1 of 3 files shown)


david-code:
@josebui Question here: an easy fix is just to allow the foreign key between `SharedData` and `User`. However I'm not sure if there if this is desired. Do you think it's OK for some data not to have an `created_by`? Or is it possible that elsewhere in the code we assume that the every data file has been created by a user?

paulschreiber (approved):

paulschreiber:
@david-code If we hard delete the user and there are files, are the files orphaned? Should they get a new owner? Perhaps the group manager? Discuss the desired behaviour with @DerekCaelin .

DerekCaelin:
My expectation is that the files would persist in the landscapes/groups where they were uploaded, and that landscape/group managers would continue to have edit/delete rights. So, the files would be orphaned, but no discernable change to users.

david-code:
> My expectation is that the files would persist in the landscapes/groups where they were uploaded, and that landscape/group managers would continue to have edit/delete rights. So, the files would be orphaned, but no discernable change to users.

Great. I'm pretty sure that is the current behaviour, but I'll add a test to make sure.

PR #240, opened 2023-01-11
feat: Create command to clean up old soft-deleted objects

### LINKED ISSUE #232: Create task to delete soft-deleted items after a fixed time period
## Description

Currently many of our models are "soft-deleted". This means instead of the data being deleted from the database, there is a column "deleted_at" set to the deletion time, and the row is ignored in queries fetching data. However we would also like to actually delete these rows after a fixed time period.

This issue will be to set up
- a recurring task
- that deletes items that were soft-deleted more than a fixed time period ago (say one month)

david-code:
## Description

Adds a new command to the `manage.py` script called `hardelete`. The command checks rows in the database having a `deleted_at` field set to more than 4 weeks ago. Matching rows are removed from the database.

Separate from this PR I will also create a cron job to run against staging. It might make sense to run this staging cron job for a while before creating a cron job for production, in case something subtly breaks.

### Checklist
- [x] New tests added

### Related Issues
References #232


terraso_backend/apps/core/management/commands/harddelete.py
```diff
@@ -0,0 +1,49 @@
...
+from safedelete.models import HARD_DELETE
+
+logger = structlog.get_logger(__name__)
+
+
+class Command(BaseCommand):
+
+    DEFAULT_DELETION_GAP = timedelta(weeks=4)
+
+    help = "Hard delete rows in database soft-deleted longer than a certain time"
```
paulschreiber, line 16:
```suggestion
    help = "Hard delete rows in database that were soft-deleted before a specified time"
```

terraso_backend/apps/core/management/commands/harddelete.py
```diff
@@ -0,0 +1,55 @@
...
+
+class Command(BaseCommand):
+
+    DEFAULT_DELETION_GAP = timedelta(days=settings.HARDDELETE_DELETION_GAP)
+
+    help = "Hard delete rows in database that were soft-deleted before a specified time"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--exec_time",
```
paulschreiber, line 21: issue: still find this name confusing.

terraso_backend/config/settings.py
```diff
@@ -236,3 +236,5 @@
 PLAUSIBLE_URL = config("PLAUSIBLE_URL", default="https://plausible.io/api/event")
 RENDER_API_URL = config("RENDER_API_URL", default="https://api.render.com/v1/")
 RENDER_API_TOKEN = config("RENDER_API_TOKEN", default="")
+
+HARDDELETE_DELETION_GAP = config("HARDDELETE_DELETION_GAP_DAYS", default="30", cast=config.eval)
```
paulschreiber, line 240: question: can we pass the date in as `30` instead of `"30"`?

paulschreiber (changes requested):

terraso_backend/apps/core/management/commands/harddelete.py
david-code: I added a help statement below to explain what this does. I'm not sure how to give this a more descriptive name, do you have any suggestions?

terraso_backend/config/settings.py
david-code: Yeah, I think so. I copied what is used elsewhere in the file but I think it would be the same result.

terraso_backend/apps/core/management/commands/harddelete.py
paulschreiber: I'm confused as to why there are two arguments. You are providing both an offset and a gap? So you can say "30 days from yesterday"?

shrouxm: i also think exec_time is a bit confusing when not viewed from the perspective of testing the code, it makes me think it's an argument for scheduling when the deletion will run. if it is not useful outside of testing, could we either manually compute the desired deletion gap in the test code or at least omit `--exec_time` from the command line arguments so it doesn't confuse someone trying to use the command?

david-code: >  if it is not useful outside of testing, could we either manually compute the desired deletion gap in the test code or at least omit --exec_time from the command line arguments so it doesn't confuse someone trying to use the command?

@paulschreiber do you think I should just remove the argument? I will have to rewrite the test, which will take additional time.

paulschreiber: @david-code Yes. I agree with my previous statement that having two arguments is confusing.

terraso_backend/apps/core/management/commands/harddelete.py
```diff
@@ -0,0 +1,54 @@
+from datetime import datetime, timedelta, timezone
```
paulschreiber, line 1: issue: addc copyright notice.

paulschreiber (changes requested):
Please remove the `--exec-time` argument.

terraso_backend/config/settings.py
```diff
@@ -369,6 +369,8 @@
     "/web-client/sitemap.xml",
 ]
 
+HARDDELETE_DELETION_GAP = config("HARDDELETE_DELETION_GAP_DAYS", default="30", cast=config.eval)
```
paulschreiber, line 372:
```suggestion
HARD_DELETE_DELETION_GAP = config("HARD_DELETE_DELETION_GAP_DAYS", default="30", cast=config.eval)
```

terraso_backend/apps/core/management/commands/harddelete.py
```diff
@@ -0,0 +1,62 @@
...
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from django.db import models
+from safedelete.models import HARD_DELETE
+
+logger = structlog.get_logger(__name__)
+
+
+class Command(BaseCommand):
+    DEFAULT_DELETION_GAP = timedelta(days=settings.HARDDELETE_DELETION_GAP)
```
paulschreiber, line 28:
```suggestion
    DEFAULT_DELETION_GAP = timedelta(days=settings.HARD_DELETE_DELETION_GAP)
```

paulschreiber (approved):

PR #269, opened 2023-02-06
feature: oauth server 246

david-code:
## Description

This is just a test I made to make sure the [Open ID Connect](https://django-oauth-toolkit.readthedocs.io/en/latest/oidc.html) endpoint, userinfo, returns the user email as expected. I am running into a small problem with Kobo, where the user's email is not being set properly in the signup form they get when they sign up with SSO. I wanted to make sure the problem was not on our side, and understand a bit deeper this endpoint itself, before I took a look at the Kobo code itself.


terraso_backend/config/urls.py
```diff
@@ -13,13 +13,65 @@
...
+            name="authorized-token-list",
+        ),
+        path(
+            "authorized-tokens/<pk>/delete/",
+            oauth2_views.AuthorizedTokenDeleteView.as_view(),
+            name="authorized-token-delete",
+        ),
+    ]
+
+
```
paulschreiber, line 68: question: Are we manually re-adding the paths we removed via `oauth2_provider.urls`? If so, is there a less duplicative way of doing this?

paulschreiber (approved):

terraso_backend/config/urls.py
david-code: Yeah, the idea is to remove certain paths that allow any authenticated user to add new OAuth applications to Terraso. I basically copied an [example](https://django-oauth-toolkit.readthedocs.io/en/latest/tutorial/tutorial_02.html) from the django-oauth-toolkit docs, so unfortunately I don't think there is an easier way.

Issue #316, opened 2023-03-06
Use django-allauth for OAuth client


david-code:
## Description

Currently we use a custom OAuth client that we have coded. It's worked well up until now, but it might be nice to use an open source library to remove the need to maintain this custom code from our codebase. [django-allauth](https://django-allauth.readthedocs.io/en/latest/) is a popular library that supports basically any SSO provider we could think of. It would make adding new SSO providers easier and less prone to bugs.

PR #317, opened 2023-03-06
feature: Add Microsoft SSO login

david-code:
## Description

Follows the pattern of the existing SSO implementation to add Microsoft as a SSO provider.

Made a [separate issue](https://github.com/techmatters/terraso-backend/issues/316) to consider using the library `django-allauth`
for SSO clients.


.env.sample
```diff
@@ -22,6 +22,11 @@ APPLE_PRIVATE_KEY=your-apple-private-key
 APPLE_CLIENT_ID=your-apple-client-id
 APPLE_AUTH_REDIRECT_URI=http://127.0.0.1:8000/auth/apple/callback
 
+MICROSOFT_CLIENT_ID=client-id
+MICROSOFT_CLIENT_SECRET=client-secret
```
paulschreiber, line 26: question: should we use a certificate?

terraso_backend/apps/auth/providers.py
```diff
@@ -149,3 +149,57 @@ def _build_client_secret(self):
             algorithm=self.JWT_ALGORITHM,
             headers=jwt_header,
         )
+
+
+class MicrosoftProvider:
+    OAUTH_BASE_URL = (
+        f"https://login.microsoft.com/{settings.MICROSOFT_TENANT}/oauth2/v2.0/authorize?"
+    )
+    TOKEN_URI = f"https://login.microsoftonline.com/{settings.MICROSOFT_TENANT}/oauth2/v2.0/token"
```
paulschreiber, line 158: suggestion: these two URLs are almost the same. combine into `MS_BASE_URL` (use a different constant name if you prefer).

terraso_backend/apps/auth/providers.py
```diff
@@ -149,3 +149,57 @@ def _build_client_secret(self):
             algorithm=self.JWT_ALGORITHM,
             headers=jwt_header,
         )
+
+
+class MicrosoftProvider:
+    OAUTH_BASE_URL = (
```
paulschreiber, line 155: question: should this be `BASE_URI` instead of `BASE_URL` to be consistent?

terraso_backend/apps/auth/services.py
```diff
@@ -62,7 +62,22 @@ def sign_up_with_apple(self, authorization_code, first_name="", last_name=""):
 
         return self._persist_user(tokens.open_id.email, first_name=first_name, last_name=last_name)
 
+    def sign_up_with_microsoft(self, authorization_code):
+        provider = MicrosoftProvider()
+        tokens = provider.fetch_auth_tokens(authorization_code)
+        if not tokens.is_valid:
+            error_msg = f"Error fetching auth tokens: {tokens.error_description}"
+            logger.error(error_msg)
```
paulschreiber, line 70: question: should this be `logger.exception`?

paulschreiber (changes requested):

terraso_backend/apps/auth/providers.py
david-code: Yes, good point. I will change this.

.env.sample
david-code: Made an issue #321 to address this. It wouldn't be a lot of new code, but it might be easier to do it in a separate PR. What do you think @paulschreiber ? Would you prefer just to do it in this PR?

terraso_backend/apps/auth/services.py
david-code: I think `error` is fine here, as no exception stacktrace needs to be printed.

terraso_backend/apps/auth/providers.py
```diff
@@ -151,11 +151,12 @@ def _build_client_secret(self):
...
+MS_BASE_URI = f"https://login.microsoft.com/{settings.MICROSOFT_TENANT}/oauth2/v2.0/"
+
+
 class MicrosoftProvider:
-    OAUTH_BASE_URL = (
-        f"https://login.microsoft.com/{settings.MICROSOFT_TENANT}/oauth2/v2.0/authorize?"
-    )
-    TOKEN_URI = f"https://login.microsoftonline.com/{settings.MICROSOFT_TENANT}/oauth2/v2.0/token"
+    OAUTH_BASE_URI = MS_BASE_URI + "authorize?"
+    TOKEN_URI = MS_BASE_URI + "token"
```
paulschreiber, line 159:
```suggestion
    OAUTH_BASE_URI = f"{MS_BASE_URI}authorize?"
    TOKEN_URI = f"{MS_BASE_URI}token"
```

.env.sample
paulschreiber: Separate PR is fine.

paulschreiber (changes requested):

paulschreiber (approved):

PR #323, opened 2023-03-13
feat: Story map url identifier

josebui:
## Description

Added story map url identifier, this removes the need to validate the uniqueness for the title

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....


paulschreiber (changes requested):
This looks good. One change requested: `url_identifier` should be changed to something more accurate/descriptive, i.e. `storymap_id` since it is identifying a storymap, not a URL.

PR #354, opened 2023-03-21
feature: email notifications

paulschreiber:
## Description
Notify users when a group membership request is pending/approved.

Add support for SES email sending
- add `django-ses`module
- add environment variables 

Add preference for notifications.

### Checklist
- [x] New tests added

### Related issues
Fixes #824,


terraso_backend/apps/core/models/users.py
```diff
@@ -114,6 +115,33 @@ def soft_delete_policy_action(self, **kwargs):
...
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}".strip()
+
+    def notifications_enabled(self):
+        preferences = self.preferences.filter(key="notifications")
+        if len(preferences) != 1:
+            return False
+
+        try:
+            return bool(preferences[0].value)
```
david-code, line 127: suggestion: the Python bool conversion only returns `False` for an empty string and `True`. If that's not what we want, we can do something like the following:

```suggestion
            return preferences[0].value.lower() == "true"
```

terraso_backend/apps/notifications/apps.py
```diff
@@ -0,0 +1,22 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see https://www.gnu.org/licenses/.
+
+from django.apps import AppConfig
+
+
+class AuthConfig(AppConfig):
+    default_auto_field = "django.db.models.BigAutoField"
+    name = "apps.notifications"
+    label = "terraso_notifications"
```
david-code, line 22: comment: This is cool, didn't know this existed!

david-code (approved):
Looks good to me!

terraso_backend/apps/core/models/users.py
paulschreiber: Not sure what we storing in the DB yet — will update to match your suggestion once that is settled.

terraso_backend/apps/notifications/apps.py
paulschreiber: Copied it from the Auth app we already have. Not sure what it does. Also fixed the class name.

terraso_backend/apps/core/models/users.py
```diff
@@ -114,6 +115,36 @@ def soft_delete_policy_action(self, **kwargs):
     def is_group_manager(self, group_id):
         return self.memberships.managers_only().filter(group__pk=group_id).exists()
 
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}".strip()
```
josebui, line 119: Suggestion: If the structure of the full name can be localized it will be more flexible for other languages in the future, the first_name last_name order is not the same in all languages

terraso_backend/apps/core/models/users.py
```diff
@@ -114,6 +115,36 @@ def soft_delete_policy_action(self, **kwargs):
...
+        return f"'{self.full_name()}' <{self.email}>"
+
+    def notifications_enabled(self):
+        preferences = self.preferences.filter(key="notifications")
+        if len(preferences) != 1:
+            return False
+
+        try:
+            return preferences[0].value.lower() == "true"
+        except AttributeError:
```
josebui, line 131: Question: When is thie expection tirggered?

terraso_backend/apps/core/models/users.py
```diff
@@ -114,6 +115,36 @@ def soft_delete_policy_action(self, **kwargs):
...
+            return False
+
+    def language(self):
+        preferences = self.preferences.filter(key="language")
+        if len(preferences) != 1:
+            return settings.LANGUAGE_CODE
```
josebui, line 137: Suggestion: `LANGUAGE_CODE` seems to be the default language code, I think `LANGUAGE_CODE_DEAFULT` or `DEFAULT_LANGUAGE_CODE` will make this easier to understand

terraso_backend/apps/core/models/users.py
```diff
@@ -114,6 +115,36 @@ def soft_delete_policy_action(self, **kwargs):
...
+
+        try:
+            language_code = preferences[0].value
+            if language_code[0:2] in [lang[0] for lang in settings.LANGUAGES]:
```
josebui, line 141: Suggestion: The system should handle locales not just languages, there can be fallbacks defined but in general just handling languages is not enough. Maybe a comment at this stage to not lose the future improvement can be enough?

terraso_backend/apps/graphql/schema/memberships.py
```diff
@@ -100,6 +101,7 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
 
         if group.membership_type == Group.MEMBERSHIP_TYPE_CLOSED:
             membership.membership_status = Membership.PENDING
+            EmailNotification.SendMembershipRequest(user, group)
```
josebui, line 104: Suggestion: Can we import and call the function directly? What is the use for the `EmailNotification` class?

josebui, line 104: Issue: function start with a lowercase character and should be snake cased: `send_membership_request `

josebui (changes requested):

terraso_backend/apps/core/models/users.py
paulschreiber: this happens if `.value` fails. It's possible I don't need the try/except block here.

terraso_backend/apps/core/models/users.py
paulschreiber: That's a good point. How do you want to handle it? Is there a module/library for this `make_full_name(first, last, locale)`?

terraso_backend/apps/core/models/users.py
paulschreiber: I agree! However, `LANGUAGE_CODE` is a special Django configuration value: https://docs.djangoproject.com/en/4.1/ref/settings/. We could duplicate that i.e.
`DEFAULT_LANGUAGE_CODE = LANGUAGE_CODE` in settings.py if you wanted to improve readibility.

terraso_backend/apps/core/models/users.py
josebui: @paulschreiber not sure in python, in the web client we are handling this with interpolations so there is a string that can define the order for each locale: https://github.com/techmatters/terraso-web-client/blob/main/src/localization/locales/es-ES.json#L13, so if there are interpolations in the library that you are using that should be enough

terraso_backend/apps/core/models/users.py
josebui: Yes, that is what I am thiking I think we should use expections for case that we can't handle in a different way, maybe check if the value exists before can be an option

terraso_backend/apps/core/models/users.py
josebui: I think reassing it to `DEFAULT_LANGUAGE_CODE = LANGUAGE_CODE` makes sense to me, but just a suggestion

terraso_backend/apps/core/models/users.py
josebui: I think it is a broader issue, I see that the po files are related to language and not locales, so eventually when we want to have more granularity here we will have to refactor it, but this can be a future improvement since the locale is in the language_code

terraso_backend/apps/graphql/schema/memberships.py
josebui: I think that if the only thing that the class is doing is proving more redability I think we should improve the name of the functions, also a suggestion I don't think we have more usages of static functions anywhere in the code, but I can be wrong

terraso_backend/apps/core/models/users.py
paulschreiber: Great. I've done the equivalent in Python.

josebui (approved):
Looks good

PR #367, opened 2023-03-27
feat: Change Microsoft SSO to use certificates

### LINKED ISSUE #321: Use client certificate encoded JWT for Microsoft SSO instead of client secret
## Description

Microsoft suggests using a [client certificate](https://learn.microsoft.com/en-us/azure/active-directory/develop/active-directory-certificate-credentials) at the ["request access token"](https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-access-token-with-a-certificate-credential) step of OAuth2. The client secret is replaced by a JWT that is encrypted with our own private certificate. As this adds an extra layer of security, it would be best to implement it this way.

david-code:
## Description

Switches Microsoft SSO to use the certificate instead of the secret key. The process is not super straightforward, so I also added a markdown file explaining how to generate a self-signed certificate.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #321

### Verification steps
- Generate self-signed certificate following docs added
- Add cert to Azure app
- Change necessary env variables
- Try signing in with a Microsoft account


terraso_backend/config/settings.py
```diff
@@ -232,6 +232,11 @@
 
 MICROSOFT_CLIENT_ID = config("MICROSOFT_CLIENT_ID", default="")
 MICROSOFT_CLIENT_SECRET = config("MICROSOFT_CLIENT_SECRET", default="")
+MICROSOFT_TENANT = config("MICROSOFT_TENANT", default="")
+MICROSOFT_PRIVATE_KEY = config("MICROSOFT_PRIVATE_KEY", default="").strip()
+MICROSOFT_CERTIFICATE_THUMBPRINT = base64.b64encode(
+    bytes.fromhex(config("MICROSOFT_CERTIFICATE_THUMBPRINT", default="").strip())
+).decode("utf-8")
```
paulschreiber, line 239: question: should this be decoded to `ascii`? It might not matter.

terraso_backend/config/settings.py
```diff
@@ -232,6 +232,11 @@
 
 MICROSOFT_CLIENT_ID = config("MICROSOFT_CLIENT_ID", default="")
 MICROSOFT_CLIENT_SECRET = config("MICROSOFT_CLIENT_SECRET", default="")
+MICROSOFT_TENANT = config("MICROSOFT_TENANT", default="")
```
paulschreiber, line 235:
```suggestion
```

paulschreiber: issue: we don't use the tenant anywhere — no need to add it.

.env.sample
```diff
@@ -24,6 +24,8 @@ APPLE_CLIENT_ID=your-apple-client-id
 MICROSOFT_CLIENT_ID=client-id
 MICROSOFT_CLIENT_SECRET=client-secret
 MICROSOFT_TENANT=tenant
```
paulschreiber, line 26:
```suggestion
```

paulschreiber (changes requested):

terraso_backend/config/settings.py
david-code: Yes, this snuck in in a rebase. Thanks for catching it.

terraso_backend/config/settings.py
david-code: Yes, I guess it could be ascii. I generally just treat strings at utf-8 unless specified otherwise, and it works fine here.

paulschreiber (approved):

PR #368, opened 2023-03-28
fix: Improve error message on missing SSO email

### LINKED ISSUE #304: feat: Handle error for SSO signup if user does not have email set
## Description

With Google and Apple SSO it seems we have a guarantee that the user will have their email set. This is good for us; the email is a required column in our user table. However, with Microsoft it seems like we do not have this guarantee, which causes us to display an ugly error message in the case where the user does not have an email defined. There are several ways we could handle this:

1) Display an error message that instructs the user to set their email. This is definitely less user friendly, but may be easier to implement.

2) Allow the user to set their email interactively. This is what is done, for example, with Kobo. The idea here would be to set up a signup page as the final signup step. This page would be pre-populated with whatever content we get from Microsoft, but it will allow the user to manually set their email if it is not present.

Any other options you see?

david-code:
## Description

If a user signs up with OAuth and the OAuth provider does not provide an email, the backend throws an inscrutable error about database constraints. This helps a bit and displays a clearer error message.

### Related Issues
Fixes #304

### DIFF
--- a/terraso_backend/apps/auth/services.py
+++ b/terraso_backend/apps/auth/services.py
@@ -77,6 +77,10 @@ class AccountService:
         )
 
     def _persist_user(self, email, first_name="", last_name="", profile_image_url=None):
+        if not email:
+            # it is possible for the email not to be set, notably with Microsoft
+            # here throw a more descriptive error message
+            raise ValueError("Could not create account, user email is empty")
         user, created = User.objects.get_or_create(email=email)
 
         self._update_profile_image(user, profile_image_url)


terraso_backend/apps/auth/services.py
```diff
@@ -77,6 +77,10 @@ def sign_up_with_microsoft(self, authorization_code):
         )
 
     def _persist_user(self, email, first_name="", last_name="", profile_image_url=None):
+        if not email:
+            # it is possible for the email not to be set, notably with Microsoft
+            # here throw a more descriptive error message
+            raise ValueError("Could not persist user, user email is empty")
```
paulschreiber, line 83: issue: Not sure what  "persist user" means. Perhaps "could not create account"?

paulschreiber (changes requested):

PR #375, opened 2023-03-30
feat: Create graphql mutation for site creation

### LINKED ISSUE #369: GraphQL mutation to create site
# Description

Add a GraphQL mutation to create a site.

# Definition of done

- [ ] GraphQL mutation to create sites available
- [ ] Mutation does not create site if required information is missing or data invalid
- [ ] Sites should support the following inputs
    - latitude
    - longitude
    - name (limit to 100 characters, no control characters - we should define what that means)
    - privacy (public or private)
- [ ] Invalid input rejected with a relevant error message

david-code:
## Description

This adds a very minimal, partial version of Site creation. There is a GraphQL endpoint that allows creating a simple site, as well as a basic Site data model.

I'd like to talk to @shrouxm and @JeancarloBarrios about the design here and iterate on it. Specifically it would be nice to think about how to improve the modularity.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #369


terraso_backend/apps/graphql/schema/sites.py
```diff
@@ -0,0 +1,31 @@
...
+
+class SiteAddMutation(BaseWriteMutation):
+    site = graphene.Field(SiteNode, required=True)
+
+    model_class = Site
+
+    class Input:
+        name = graphene.String(required=True)
+        lat_deg = graphene.Float(required=True)
+        lon_deg = graphene.Float(required=True)
```
paulschreiber, line 31: Suggestion: rename this to `latitude` (or `lat`) and `longitude` (or `lng`). You could also pick the long name and alias the short name as a function. We use `lat` an`lng` elsewhere in the code.

terraso_backend/apps/graphql/schema/sites.py
```diff
@@ -0,0 +1,31 @@
+import graphene
```
paulschreiber, line 1: issue: add a copyright notice.

terraso_backend/apps/soilproj/__init__.py
paulschreiber, line 1: suggestion: rename `soliproj` to `soilid` (assuming this is for Soil ID).

terraso_backend/apps/soilproj/models.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Site(SlugModel):
+    class Meta(SlugModel.Meta):
+        abstract = False
+        _unique_fields = ["name"]
+
+    name = models.CharField(max_length=200)
+    lat_deg = models.FloatField()
+    lon_deg = models.FloatField()
```
paulschreiber, line 27: issue: same naming suggestion as above.

paulschreiber (changes requested):

terraso_backend/apps/soilproj/models.py
```diff
@@ -0,0 +1,29 @@
...
+
+    field_to_slug = "name"
```
shrouxm, line 29: suggestion: can we use ID as slug? or something other than name. see discussion here: https://github.com/techmatters/terraso-web-client/issues/751

terraso_backend/apps/soilproj/models.py
```diff
@@ -0,0 +1,29 @@
...
+# along with this program. If not, see https://www.gnu.org/licenses/.
+from django.db import models
+
+from apps.core.models.commons import SlugModel
+
+
+class Site(SlugModel):
+    class Meta(SlugModel.Meta):
+        abstract = False
+        _unique_fields = ["name"]
```
shrouxm, line 23: suggestion: if we aren't using name as slug then it needn't be globally unique

terraso_backend/apps/soilproj/migrations/0001_initial.py
```diff
@@ -0,0 +1,58 @@
...
+    initial = True
+
+    dependencies = []
+
+    operations = [
+        migrations.CreateModel(
+            name="Site",
+            fields=[
+                ("deleted_at", models.DateTimeField(db_index=True, editable=False, null=True)),
+                ("deleted_by_cascade", models.BooleanField(default=False, editable=False)),
```
shrouxm, line 19: question: what is this meant to track? like if it gets deleted because it's part of a project that was deleted?

terraso_backend/tests/graphql/test_sites.py
```diff
@@ -0,0 +1,46 @@
...
+import json
+
+import pytest
+
+from apps.soilproj.models import Site
+
+pytestmark = pytest.mark.django_db
+
+
+def test_site_creation(client_query):
```
shrouxm, line 24: suggestion: if we are using a method of generating slugs that does not guarantee uniqueness, a test to make sure that duplicate slugs are rejected

shrouxm (commented):
i still feel too beginner with django to be able to comment on the modularity, but nothing jumps out at me as a red flag there

terraso_backend/apps/graphql/schema/sites.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class SiteAddMutation(BaseWriteMutation):
+    site = graphene.Field(SiteNode, required=True)
+
+    model_class = Site
+
+    class Input:
+        name = graphene.String(required=True)
+        lat_deg = graphene.Float(required=True)
```
JeancarloBarrios, line 30:
```suggestion
        lat = graphene.Float(required=True)
```

terraso_backend/apps/graphql/schema/sites.py
JeancarloBarrios, line 31:
```suggestion
        lat = graphene.Float(required=True)
        lng = graphene.Float(required=True)
```

JeancarloBarrios: +1

terraso_backend/apps/soilproj/models.py
JeancarloBarrios, line 27:
```suggestion
    lat = models.FloatField()
    lng = models.FloatField()
```

terraso_backend/apps/graphql/schema/sites.py
david-code: I actually rely prefer having the units very explicitly included in the variable name. This helps avoid logic errors.

terraso_backend/apps/soilproj/__init__.py
david-code: I was thinking of keeping this Django app focused on just the project management. We would probably make a different project for running the Soil ID algorithm, for example.

terraso_backend/apps/soilproj/migrations/0001_initial.py
david-code: This field is generated by a configuration setting we have in [django-safedelete](https://django-safedelete.readthedocs.io/en/latest/index.html?highlight=safe_delete_field_name#configuration), a Django plugin we use to support "soft" deleting.

The idea is that when we delete a model, the field `deleted_at` is set to the date it was deleted. The plugin modifies the ORM Manager class so that the SELECT queries ignore any rows with this column NOT NULL.

`deleted_by_cascade` is used as well by the plugin to know if the model should be "undeleted" when a linked model is undeleted. So maybe if a site was deleted because the related project was deleted; then if the project is restored, the site should be restored as well. (I'm not sure if that's actually how project deletion will work, but it's an example of what the field is used for).

terraso_backend/apps/graphql/schema/sites.py
paulschreiber: This is nonstandard and will be confusing/resulting in errors. Other code (and other APIs) use lat/lng.

paulschreiber: This is nonstandard and will be confusing/resulting in errors. Other code (and other APIs) use lat/lng.

terraso_backend/apps/soilproj/migrations/0001_initial.py
shrouxm: cool, thank you for the explanation! django scares me

terraso_backend/apps/soilproj/__init__.py
shrouxm: imo: if there won't be any code in this django app related to Soil ID (e.g. we've talked about sites and projects being useful concepts outside of LandPKS), then it shouldn't have soil in the name at all.

terraso_backend/tests/graphql/test_sites.py
david-code: @shrouxm I switched to use the ID for slugs for now. I think we can work out the details of how we should generate slugs as it becomes more relevant.

terraso_backend/apps/soilproj/__init__.py
david-code: That makes sense Garo. I wasn't sure if we wanted to commit to a concept of a project outside of soil-related stuff. I'll rename the app to `project_management`

terraso_backend/apps/project_management/apps.py
```diff
@@ -0,0 +1,20 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see https://www.gnu.org/licenses/.
+from django.apps import AppConfig
+
+
+class SoilprojConfig(AppConfig):
```
shrouxm, line 18:
```suggestion
class ProjectManagementConfig(AppConfig):
```

docker-compose.dev.yml
```diff
@@ -10,11 +10,6 @@ services:
     depends_on:
       db:
         condition: service_healthy
-
-    depends_on:
```
shrouxm, line 14: question: was this just a redundant clause?

terraso_backend/apps/project_management/migrations/0001_initial.py
```diff
@@ -0,0 +1,58 @@
...
+                ),
+                ("created_at", models.DateTimeField(auto_now_add=True)),
+                ("updated_at", models.DateTimeField(auto_now=True)),
+                ("slug", models.SlugField(blank=True, editable=False, max_length=250)),
+                ("name", models.CharField(max_length=200)),
+                ("latitude", models.FloatField()),
+                ("longitude", models.FloatField()),
+            ],
+            options={
+                "ordering": ["created_at"],
```
shrouxm, line 34: question: does this create a SQL index on "created_at"? i guess this is default behavior from a parent class?

shrouxm (changes requested):
LGTM other than the one rename

docker-compose.dev.yml
paulschreiber: Yes. Was a blown merge somewhere.

paulschreiber (approved):

terraso_backend/apps/project_management/migrations/0001_initial.py
david-code: > does this create a SQL index on "created_at"?

No idea, would be interesting to check it out.

> i guess this is default behavior from a parent class?

Yes, I think I saw you mention this in Slack, it seems to be the case.

shrouxm (approved):

PR #376, opened 2023-03-31
feat: added protocol and implementation example

JeancarloBarrios:
## Description
Simple example, with some caveats but we can iterate on it

### DIFF
--- a/terraso_backend/apps/soilproj/service.py
+++ b/terraso_backend/apps/soilproj/service.py
@@ -1,0 +1,12 @@
+
+from typing import Optional
+
+from . import models
+
+
+class SoilService():
+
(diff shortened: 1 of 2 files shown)


david-code:
Finally got around to looking at this @JeancarloBarrios , looks nice! I'll try implementing this in the PR.

terraso_backend/apps/soilproj/service.py
```diff
@@ -0,0 +1,12 @@
+
+from typing import Optional
+
+from . import models
+
+
+class SoilService():
+
+    def get_site_by_name(self, name: str) -> Optional[models.Site]:
```
josebui, line 9: Question: Can you explain what is ` -> Optional[models.Site]` doing here? I am not familiar with this notation

PR #378, opened 2023-03-31
feat: Add GraphQL mutation to create new Project

### LINKED ISSUE #374: Project - Minimal Implementation
## Description

This issue is to create a very basic first version of the Project, with the following criteria:

- User can create Project
- Sites can be added to a Project

As a result, all Users will be Managers, Sites cannot be removed from Projects, etc.

## Definition of Done

- GraphQL mutation to create project
- GraphQL mutation to add site to project

david-code:
## Description

This is a very minimal implementation that creates a Project. The User creating the Project is added to the Project as well, with the rank of "Manager". It also allows managers to add sites to the project. However, project permissions, etc. have not been implemented yet, so user actions are not yet controlled properly (or even implemented).

### Implementation

The implementation follows how groups and landscapes are implemented in Terraso. When a project is created, a new default group is also created and linked to the project. This default group is used to control adding users to the project, managing their permission, removing them from the project, etc.

image

I chose to implement this as a 1-1 relationship, instead of the many-to-many relationship that exists currently between landscapes and groups. I think it makes sense that each project only has a single group.

### Group Membership Changes

After some deliberation, I realized that two issues could be spit up:
1) Users joining projects: users cannot join projects, they can only be added by project manager. This means that they should not be able to join a group associated with a project.
2) Users accessing sites in  projects: this depends on the "public" and "private" settings, and can managed at the project level.

Realizing this, I added a new type of group membership, "restricted". This type of group only allows users to join if they are being added by a manager. I don't think this solution is perfect, but I think it works pretty well.

### Additional changes

The GraphQL queries for groups and landscapes will need to be updated slightly. I've chosen to just merge this PR into a feature branch, and just create new PR with these changes. The changes are fairly minor, but there is already enough to review in this PR as it is.

### Checklist
- [x] Corresponding issue has been opened (#374)
- [x] New tests added

### Related Issues
Close #374


terraso_backend/apps/soilproj/models.py
```diff
@@ -0,0 +1,73 @@
...
+
+        rules_permissions = {"add_site": permission_rules.can_add_site}
+
+    PRIVATE = "pri"
+    PUBLIC = "pub"
+    PRIVACY_OPTIONS = [(PRIVATE, "Private"), (PUBLIC, "Public")]
+
+    name = models.CharField(max_length=200)
+    privacy = models.CharField(max_length=3, choices=PRIVACY_OPTIONS, default=PRIVATE)
+    members = models.ManyToManyField(User, through="ProjectMembership")
```
shrouxm, line 34: question: there was discussion of having the project contain a group in the way that landscapes do to manage membership, do you have thoughts on this? i'm not sure what the pros and cons would be. like the vague pro would be that it might make integrating LandPKS with Terraso's other features easier, & the vague con would be that it would couple projects to groups in a potentially annoying way in the future.

david-code: Interesting thought! I am not against this but it would make sense to think it out a bit further before we try implementing this.

terraso_backend/apps/core/permission_rules.py
```diff
@@ -103,6 +103,16 @@ def allowed_to_update_preferences(user, user_preferences):
     return user_preferences.user.id == user.id
 
 
+@rules.predicate
+def is_site_creator(user, site):
```
josebui, line 107: Suggestion: Follow the same naming convention as the other rules `allowed_to_...`

terraso_backend/apps/soilproj/admin.py
```diff
@@ -0,0 +1,17 @@
+# Copyright © 2021-2023 Technology Matters
```
josebui, line 1: Suggestion: The name `soilproj` is not understandable, can we have a complete name here

terraso_backend/apps/soilproj/models.py
```diff
@@ -0,0 +1,73 @@
+# Copyright © 2021-2023 Technology Matters
```
josebui, line 1: Issue: The models in this file are missing the `rules_permissions` definition, see: https://github.com/techmatters/terraso-backend/blob/main/terraso_backend/apps/core/models/groups.py#L84

josebui (changes requested):

terraso_backend/apps/graphql/schema/projects.py
```diff
@@ -0,0 +1,80 @@
+# Copyright © 2021-2023 Technology Matters
```
josebui, line 1: Suggestion: I think we need a new subfolder under `/graphql/schema` `site` and `project` are not self explainatory without knowing that is related to soild ID

terraso_backend/apps/graphql/schema/projects.py
```diff
@@ -0,0 +1,80 @@
...
+    class Input:
+        siteID = graphene.ID(required=True)
+        projectID = graphene.ID(required=True)
+
+    @classmethod
+    def mutate_and_get_payload(cls, root, info, projectID, siteID):
+        site = Site.objects.get(id=siteID)
+        project = Project.objects.get(id=projectID)
+        user = info.context.user
+        if not user.has_perm(Site.get_perm("add_to_project"), site) or not user.has_perm(
```
josebui, line 74: Suggestion: Maybe using `rules_permissions` you don't need to customize this validation

terraso_backend/apps/graphql/schema/sites.py
```diff
@@ -0,0 +1,37 @@
...
+        model = Site
+
+        filter_fields = {"name": ["icontains"]}
+        fields = ("name", "lat_deg", "lon_deg")
+
+        interfaces = (relay.Node,)
+        connection_class = TerrasoConnection
+
+
+class SiteAddMutation(BaseWriteMutation):
```
josebui, line 23: Suggestion: I see some mutation inheriting from `BaseMutation` and others from `BaseWriteMutation` we should try to use `BaseWriteMutation` as much as possible

terraso_backend/apps/soilproj/models.py
josebui: I agree with @shrouxm that is was the expectation since the group has all the logic to handle memberships, the idea was to implement this similarly to landscape

david-code: I think you're right @josebui , we need to think through how these concepts will interact. I'm working on an architecture issue for this right now, I'll try adding my thoughts/questions there.

terraso_backend/apps/soilproj/models.py
david-code: I think they have `rules_permissions`, but I noticed reading the docs a bit more that if I use `change`, `delete`, etc. this will be handled in the model validation, is that right?

josebui: @david-code As far as I understand, this is the model validation, or what do you mean by model validation?

terraso_backend/apps/graphql/schema/projects.py
david-code: @josebui is this still relevant?

terraso_backend/apps/core/models/groups.py
```diff
@@ -102,9 +105,47 @@ def save(self, *args, **kwargs):
...
+    def _add_user(
+        self,
+        user: User,
+        role: Union[Literal["manager"], Literal["member"]],
+    ):
+        self.memberships.update_or_create(group=self, user=user, defaults={"user_role": role})
+
+    @property
+    def group_managers(self):
+        manager_memberships = models.Subquery(self.memberships.managers_only().values("user_id"))
```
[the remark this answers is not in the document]
david-code: Yeah I had to same thought, but managers_only only includes memberships that are approved.

david-code:
@josebui I was thinking about what will need to be done to filter restricted groups from the frontend. If I understand it correctly, we will just need to add a new GraphQL filter on the backend that will let the frontend query filter out restricted groups. Is there perhaps a different way you would suggest?

terraso_backend/apps/core/models/groups.py
```diff
@@ -40,11 +41,13 @@ class Group(SlugModel):
 
     MEMBERSHIP_TYPE_OPEN = "open"
     MEMBERSHIP_TYPE_CLOSED = "closed"
+    MEMBERSHIP_TYPE_RESTRICTED = "restricted"
```
josebui, line 44: Suggestion: From what I understand it seems that `restricted` is not the type of the group, it is the mechanism available to join the group, and that can have different variations:
1. User can join a group
2. User can be added to a group
3. User can join or be added to a group

From the conversation we had the other day, what we want to have in the long term is a data entity that is separated from group, landscape and project that handles the logic related to memberships. So, from that I understand that maybe having all of that now is a big change, but I think we should move a step closer to that long term goal, and from that point of view, we need a way to define how the user can join the group, maybe we can have something called `join_method`?

terraso_backend/apps/project_management/models/projects.py
```diff
@@ -0,0 +1,59 @@
...
+        rules_permissions = {"change": permission_rules.allowed_to_change_project}
+
+    PRIVATE = "private"
+    PUBLIC = "public"
+    DEFAULT_PRIVACY_STATUS = PRIVATE
+
+    PRIVACY_STATUS = ((PRIVATE, _("Private")), (PUBLIC, _("Public")))
+
+    name = models.CharField(max_length=200)
+    group = models.OneToOneField(Group, on_delete=models.CASCADE)
```
josebui, line 36: Suggestion: I think we can also start detaching this from the group definition, can we name it `membership` or something like that?

terraso_backend/apps/project_management/models/sites.py
```diff
@@ -14,15 +14,32 @@
...
+
     name = models.CharField(max_length=200)
+    name = models.CharField(max_length=200)
+    latitude = models.FloatField()
     latitude = models.FloatField()
     longitude = models.FloatField()
+    longitude = models.FloatField()
 
     field_to_slug = "id"
+    field_to_slug = "id"
```
josebui, line 37: Question: If the slug is not based on a human redable name, do we need to use SlugModel?

terraso_backend/apps/core/models/groups.py
josebui, line 44: Question: What will happen with the current queries for groups? will the project groups return? I think maybe what is more relevant right now (if we are not able to create the separate memberhip entity) is to have a `hidden` type for the groups that should not appear, but I think this is the biggest concern on not having the separate entity as part of this implementation, since `hidden` is not that clean in terms of the meaning

terraso_backend/apps/graphql/schema/projects.py
josebui: I think we are good for now, I just think we should be careful about this, what the naming of this entities is saying is that they are not always related to Soil ID, we can have projects separated from soild ID for othe objectives, I think we should have that in mind when creating the other entities, and implementing other processes

josebui:
> @josebui I was thinking about what will need to be done to filter restricted groups from the frontend. If I understand it correctly, we will just need to add a new GraphQL filter on the backend that will let the frontend query filter out restricted groups. Is there perhaps a different way you would suggest?
@david-code In one of the comments I mention a new `hidden` or maybe `memberhips_group` group type, as a work around, I still think that the good solution for this will be to have the separate membership entity, but one step closer will be to not use group type for `retricted` and add a new type for groups that are just to hold memberhips for other entities, tha way we can filter them by that type

terraso_backend/apps/project_management/models/projects.py
```diff
@@ -0,0 +1,59 @@
...
+    privacy = models.CharField(
```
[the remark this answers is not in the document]
david-code: Can you elaborate a bit Jose? I don't think I understand.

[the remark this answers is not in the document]
shrouxm: @josebui i think the distinction is that `open`, `closed`, and `restricted` are essentially edit permissions of the group membership (who can add a new member), whereas `private`/`public` is a view permission for non-members of the group. so if we keep `open`, `closed`, etc at the membership group level we can avoid duplicating the associated auth logic

shrouxm (commented):
i haven't finished reviewing the tests yet but not sure when i will get to that so leaving the comments i have. i think it makes sense to separate the refactor we want to do to split group from group membership from this PR, and to do this PR first since it's already written up, but do the refactor soon after, so i am ok with the little code smells resulting from that intermediate mess.

terraso_backend/apps/graphql/schema/projects.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class ProjectNode(DjangoObjectType):
+    id = graphene.ID(source="pk", required=True)
+
+    class Meta:
+        model = Project
+
+        filter_fields = {"name": ["icontains"]}
+        fields = ("name",)
```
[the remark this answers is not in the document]
david-code: Yeah, good point. There's a separate issue to add the queries and such which I haven't actually done yet, so I will take of that there I think.

terraso_backend/apps/graphql/schema/commons.py
```diff
@@ -147,6 +167,12 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
     def is_update(cls, data):
         return "id" in data
 
+    @classmethod
+    def not_allowed(cls, mutation_type=None):
```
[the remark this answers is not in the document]
david-code: Oh good point. I think because `BaseAuthenticatedMutation` has `model_class` property, while the other one does not. I think I could at least call the parent method from the child...

terraso_backend/apps/core/models/groups.py
```diff
@@ -102,9 +105,47 @@ def save(self, *args, **kwargs):
...
+
+    def is_member(self, user: User) -> bool:
+        return self.group_members.filter(id=user.id).exists()
+
+    @property
+    def is_restricted(self):
+        return self.membership_type == self.MEMBERSHIP_TYPE_RESTRICTED
+
+    @classmethod
+    def create_default_group_project(cls, name: str):
```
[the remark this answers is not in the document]
david-code: Good point Garo, I think this could be refactored out.

terraso_backend/apps/core/models/groups.py
david-code: > my understanding of the code is that as it is, if we make project groups "open" the previous API surface would allow manager or user to add a membership.

Yep, correct. I think this would almost work if we set groups closed, but the way it works right now is that the managers of the group are sent an email asking to confirm membership after the user is added (you can check out the `EditMembershipMutation` to see the logic). Unfortunately I think that would be really annoying for the manager to have to go to their email to confirm users they are adding.

terraso_backend/apps/project_management/models/sites.py
```diff
@@ -14,15 +14,32 @@
...
         abstract = False
 
+        rules_permissions = {"change": permission_rules.allowed_to_edit_site}
+
     name = models.CharField(max_length=200)
+    name = models.CharField(max_length=200)
+    latitude = models.FloatField()
     latitude = models.FloatField()
     longitude = models.FloatField()
+    longitude = models.FloatField()
```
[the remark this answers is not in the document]
david-code: Yes indeed, I am somewhat shocked that this does not cause things to fail.

terraso_backend/apps/core/models/groups.py
shrouxm: hmm OK, what if we just make them "open" for now and then we will close that hole up with the refactor long before we actually release anything?

josebui: @david-code I think that ideally the type of the group should not have anything to do with how the members join the group, type is more related to the data related to a group, and specific requirements linked in our case to 1000L group, landscape and project, right now it is working with closed because there were no other use cases, but now we have enough requirements to make this distiction, so I think we should remove the `restricted` type and just add a new field that will define the join behavior `join_method` that way we dont have to define a new name for a type everytime we need a new combination between (access_data(type?), user_can_join, can_invite_user)

terraso_backend/apps/project_management/models/projects.py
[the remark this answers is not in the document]
josebui: @shrouxm The requiement for `closed` and `open` groups orginated from the need to make data within a group not accessible by non-members, so this was not related to how a user can join, both right now can be joined directly or joined direclty pending for an approval.

Also, if we keep `closed` and `restricted` as group type that distinction from the name is not clear, so those names by themselve are not telling us a difference, because IMO they are being used to group different sets of requirements, if we split them (see my reply in the previous comment) we can make this more flexible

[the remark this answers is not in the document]
josebui: @david-code What I mean (sorry I am repeating a little what was said in the previous comment) is that this concepts of private, public, closed and open are overlapping through common code, and that is confusing since we are planning to split the membership logic we have to identify what is relevant to 1000L group, landspace and project, and what is relevant to the membership entity. How the data within a landscape, 1000L group or project is accesed seems like logic not related to membership, from this example even the naming from the bussiness point of view changes, 1000L groups can be open or close and projects and be public or private, so this concept IMO should not be part of the membership logic

terraso_backend/apps/core/models/groups.py
josebui: @shrouxm I don't really like our current solution, but what is happening is that the graph QL query from the client is explicilty filtering those, with the creation of a separate memberhips entity this will also work better I think. See: https://github.com/techmatters/terraso-web-client/blob/main/src/group/groupService.js#L116

terraso_backend/apps/project_management/models/projects.py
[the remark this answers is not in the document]
david-code: Got it Jose, thanks for clarifying!

terraso_backend/apps/core/migrations/0041_landscapedefaultgroup_alter_group_deleted_at_and_more.py
```diff
@@ -0,0 +1,35 @@
...
+    dependencies = [
+        ("core", "0040_website_length"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="group",
+            name="membership_type",
+            field=models.CharField(
+                choices=[("open", "Open"), ("closed", "Closed"), ("restricted", "Restricted")],
```
josebui, line 30: Issue: From the current changes in the model it seems this is not needed anymore

terraso_backend/apps/core/migrations/0042_landscapedefaultgroup_group_enroll_method_and_more.py
```diff
@@ -0,0 +1,28 @@
...
+        migrations.AddField(
+            model_name="group",
+            name="enroll_method",
+            field=models.CharField(
+                choices=[("join", "Join"), ("invite", "Invite"), ("both", "Both")],
+                default="join",
+                max_length=10,
+            ),
+        ),
+        migrations.AlterField(
```
josebui, line 21: Question: Also this one, were these migrations executed on staging? if that is the case then maybe we do need them.

terraso_backend/apps/project_management/models/projects.py
```diff
@@ -0,0 +1,68 @@
...
+from apps.core import permission_rules
+from apps.core.models import Group, User
+from apps.core.models.commons import BaseModel
+
+
+class Project(BaseModel):
+    class Meta(BaseModel.Meta):
+        abstract = False
+
+        rules_permissions = {"change": permission_rules.allowed_to_change_project}
```
josebui, line 27: Question: Do we need `delete` here?

terraso_backend/apps/project_management/models/sites.py
```diff
@@ -14,15 +14,28 @@
...
 from apps.core.models.commons import SlugModel
 
+from .projects import Project
+
 
 class Site(SlugModel):
     class Meta(SlugModel.Meta):
         abstract = False
 
+        rules_permissions = {"change": permission_rules.allowed_to_edit_site}
```
josebui, line 27: Question: Do we need `delete` here too?

terraso_backend/apps/graphql/schema/groups.py
```diff
@@ -64,6 +64,10 @@ class GroupNode(DjangoObjectType):
     account_membership = graphene.Field("apps.graphql.schema.memberships.MembershipNode")
     memberships_count = graphene.Int()
 
+    @classmethod
+    def get_queryset(cls, queryset, info):
+        return queryset.filter(project__isnull=True)
```
josebui, line 69: Suggestion: Can you add a comment here? explaining why this is being filtered

josebui:
> @josebui Two changes:
Added an enroll_method field similar to the join_method you suggested above. Let me know if this is what you were thinking. The default is meant to match the current "join" method for landscapes and 1000L groups.
Filtered out the project groups from the graphql queries, which was relatively painless (https://github.com/techmatters/terraso-backend/commit/d3476348ffdeadc58c2807f4e1e1e6fd33aff6d3). Is there anywhere else that you can think of where this needs to be done? From my perspective that seems to be it.

@david-code Yes, that looks good David, that will make the future refactor easier, thanks for that change.

About the queryset for groups I think that is good for now, I just added a suggestion to add a comment there

terraso_backend/apps/core/migrations/0041_landscapedefaultgroup_alter_group_deleted_at_and_more.py
david-code: Yes, will fix this.

terraso_backend/apps/core/migrations/0042_landscapedefaultgroup_group_enroll_method_and_more.py
david-code: No I don't think so, I haven't deployed to staging yet.

terraso_backend/apps/project_management/models/projects.py
david-code: Yes, but I will leave that for when we implement the feature to delete the project.

PR #379, opened 2023-03-31
feat: Public view support

josebui:
## Description

- Added optional authentication decorator
- Added settings list for public base paths
- Made GraphQL endpoints optionally authenticated
- Improved JWT validation to raise exceptions
- Fixed Graph QL filters using user data to handle anonymous users

### Checklist
- [x] New tests added

### Related Issues
Fixes https://github.com/techmatters/terraso-web-client/issues/775

### DIFF
--- a/terraso_backend/apps/auth/middleware.py
+++ b/terraso_backend/apps/auth/middleware.py
@@ -16,6 +16,7 @@
 from functools import wraps
 
 import structlog
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ImproperlyConfigured, ValidationError
 from django.http.response import JsonResponse
@@ -30,9 +31,9 @@ User = get_user_model()
 class JWTAuthenticationMiddleware:
     def process_view(self, request, view_func, view_args, view_kwargs):
         auth_optional = getattr(view_func, "auth_optional", False)
-        auth_required = not getattr(view_func, "auth_not_required", False) and not auth_optional
+        auth_required = not self._is_path_public(request.path) and not auth_optional
 
-        if not auth_required and not auth_optional:
+        if request.user.is_authenticated or (not auth_required and not auth_optional):
             return None
 
         try:
@@ -46,6 +47,12 @@ class JWTAuthenticationMiddleware:
 
         if auth_required:
             return JsonResponse({"error": "Unauthorized request"}, status=401)
+
+    def _is_path_public(self, path):
+        for public_path in settings.PUBLIC_BASE_PATHS:
+            if path.startswith(public_path):
+                return True
+        return False
 
     def __init__(self, get_response):
         self.get_response = get_response
@@ -92,14 +99,6 @@ class JWTAuthenticationMiddleware:
         except User.DoesNotExist:
             logger.error("User from JWT token not found", extra={"user_id": user_id})
             return None
-
-
-def auth_not_required(view_func):
-    def wrapped_view(request, *args, **kwargs):
-        return view_func(request, *args, **kwargs)
-
-    view_func.auth_not_required = True
-    return wraps(view_func)(wrapped_view)
 
 
 def auth_optional(view_func):
--- a/terraso_backend/apps/auth/middleware.py
+++ b/terraso_backend/apps/auth/middleware.py
@@ -13,6 +13,8 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see https://www.gnu.org/licenses/.
 
+from functools import wraps
+
 import structlog
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ImproperlyConfigured, ValidationError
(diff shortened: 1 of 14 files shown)


terraso_backend/apps/auth/middleware.py
```diff
@@ -23,56 +29,82 @@
 
 
 class JWTAuthenticationMiddleware:
-    def __init__(self, get_response):
-        self.get_response = get_response
+    def process_view(self, request, view_func, view_args, view_kwargs):
+        auth_optional = getattr(view_func, "auth_optional", False)
+        auth_required = not self._is_path_public(request.path) and not auth_optional
```
paulschreiber, line 34: question: does Python do short-circuit conditionals? If so, put `not auth_optional` first, since it'll be quicker.

paulschreiber (approved):

terraso_backend/apps/graphql/schema/commons.py
```diff
@@ -49,6 +51,15 @@ class Meta:
...
     def mutate(cls, root, info, input):
+        user = info.context.user
+
+        if not user or not user.is_authenticated:
+            message = {
+                "message": "You must be authenticated to perform this operation",
+                "code": "unauthorized",
+            }
+            return cls(errors=[{"message": json.dumps([message])}])
+
```
david-code, line 62: suggestion: Would this be better implemented as [middleware](https://docs.graphene-python.org/en/latest/execution/middleware/) ([graphene-django](https://docs.graphene-python.org/projects/django/en/latest/settings/#middleware))? 

I was having some problems with this function yesterday (#394 , still have to fill out the details). I think the functionality it tries to perform would be better implemented with middleware. The class hierarchy it relies on in `graphene-django` is not documented and it's difficult for me to figure out why it's not working as expected.

This is kind of a big ask for this PR, so I'd be OK with adding a comment somewhere someone adding a mutation will see it (i.e. `schemas/__init__.py`, above the `Mutations` class) explaining that all Mutations MUST extend this base class. I say this because I was considering bypassing this base class yesterday in some of the LandPKS code I was writing, and that would not be good.

terraso_backend/apps/auth/middleware.py
```diff
@@ -23,56 +29,82 @@
...
 
-    def __call__(self, request):
-        if not request.user or not request.user.is_authenticated:
-            user = self._get_user_from_jwt(request)
+        if request.user.is_authenticated or (not auth_required and not auth_optional):
+            return None
+
+        auth_header = request.META.get("HTTP_AUTHORIZATION")
+        if not auth_header and auth_optional:
+            request.user = None
```
david-code, line 41: question: Could we replace this with [AnonymousUser](https://docs.djangoproject.com/en/4.2/topics/auth/default/#authentication-in-web-requests)? I'm still not super familiar with how our current authentication setup works, but it seems like `AnonymousUser` is what is used in Django when the user is not logged in. I'm just wondering if some Django plugins might rely on `AnonymousUser` (I don't think any that we use currently, but maybe in the future).

david-code (changes requested):

terraso_backend/apps/graphql/schema/commons.py
josebui: @david-code I would like to get more context about this. This right now is just related to mutations, so this validation is inside the Graph QL schema definition, right now it is working for all mutations, but we don't know if in the future we want some types of mutations that could be accessed anonymously, still it seems that a validation at the mutation level is something that could be done. So if the issue is being able to identify from where the error generates I think that is more related to our logs, because even if we move this validation outside of the mutations, you can defenitely add validations inside mutations. 

I think the comment where we have the schame defined saying that all mutations should use it is a good thing, I am not sure if there is a way to enforce it, I was thinking about something like that, like creating custom rules

terraso_backend/apps/auth/urls.py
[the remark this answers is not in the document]
josebui: @david-code You comment doesn't show a specific liine in the code, not sure why, can you give me more context about the difference that you see

terraso_backend/apps/auth/middleware.py
josebui: I thought that was happening automatically but it seems that is not the case, I will check how to use AnonymousUser

josebui: @david-code I updated it to use AnonymousUser

terraso_backend/apps/auth/urls.py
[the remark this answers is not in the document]
david-code: It looks like there was no change to the code, and yet the formatting has changed. The only way that makes sense to me is if your local black setup has a shorter line length than the CI enforces.

terraso_backend/apps/graphql/schema/commons.py
david-code: I see, sorry, I wasn't very clear. When I said "I'm not sure why it isn't working as expected", I meant that I had problems yesterday working with this function (without your changes). It doesn't seem to correctly return the error message I expect. The rest of the comment was just noting that we are using a feature that isn't documented by `graphene-django`, namely that `mutate_and_get_payload` is called by the `mutate` method in the parent class here. I found this confusing yesterday when I was working with the code.

Instead of overriding this method to try and perform authentication checks, I was thinking of basically doing the same thing in some middleware. This would have the benefit that it would automatically apply to any mutation created in the future, without needing to extend this base class. 

However, I realize that is a bit of a complicated process, and after all I'm not sure if you agree with me that the solution is optimal. In that case, I think just adding a comment to the `__init__` file is fine; if you want to add a new mutation, you need to read that file anyways. This would just avoid someone writing a Mutation that is accidentally exposed to unauthenticated users by not extending the base class.

terraso_backend/apps/auth/urls.py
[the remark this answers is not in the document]
josebui: @david-code Ohh, I see initially there were changes, I am thinknig that at some point I manually changed to how it looks in this PR, so maybe for black there is no just one acceptable way, I will put this back to the previous format

PR #400, opened 2023-04-05
fix: simplify story map ID generation

paulschreiber:
## Description
Simplify story map ID generation by using `secrets`.

### DIFF
--- a/terraso_backend/apps/story_map/migrations/0004_remove_storymap_story_map_storymap_unique_active_slug_and_more.py
+++ b/terraso_backend/apps/story_map/migrations/0004_remove_storymap_story_map_storymap_unique_active_slug_and_more.py
@@ -13,7 +13,8 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see https://www.gnu.org/licenses/.
 
-import secrets
+import random
+import string
 
 from django.db import migrations, models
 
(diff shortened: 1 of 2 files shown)


terraso_backend/apps/story_map/migrations/0004_remove_storymap_story_map_storymap_unique_active_slug_and_more.py
```diff
@@ -13,16 +13,15 @@
...
+import secrets
 
 from django.db import migrations, models
 
 
 def update_rows(apps, schema_editor):
     StoryMap = apps.get_model("story_map", "StoryMap")
     for obj in StoryMap.objects.all():
-        obj.story_map_id = "".join(random.choices(string.ascii_lowercase + string.digits, k=7))
+        obj.story_map_id = secrets.token_hex(4)
```
josebui, line 24: Issue: This migration was already executed, we should not change it.
Question: Do we want to migrate the current IDs? if that is the case we need a separate migration for this

josebui (changes requested):

terraso_backend/apps/story_map/migrations/0004_remove_storymap_story_map_storymap_unique_active_slug_and_more.py
paulschreiber: The idea was if someone else were to check out and run Terraso, they'd get the nicer IDs. I don't think we need to migrate existing IDs.

josebui: I think that will only be valid if for some reason they decided to execute a subset of migrations, then create story maps and then execute the rest of the migrations, because at the first execution there are no story maps to populate with this, and likely the tool will not work if the execute just a subset of migrations, so I think this is more relevant to keep the current state of the DBs consistent with the migrations they have used

josebui (approved):

PR #422, opened 2023-04-14
chore: add migrate and collectstatic to Dockerfile

paulschreiber:
## Description
Add migrate and collectstatic commands to Dockerfile. Should trigger updates on deploy.


david-code:
With these changes, when the Docker image is being built the database migrations and copying the static files to S3 will be executed. It would be better to do this when the container itself is being started.

Looking at the Render settings, there is a field to set the startup command of the container. If we affixed "python terraso_backend/manage.py migrate && python terraso_backend/manage.py migrate &&" to the start of the command that we currently have, that might be able to achieve this goal better.

paulschreiber:
Render support recommended using the Dockerfile. What is the advantage of moving this work to the startup command?

david-code:
I think the major drawback is that imposes a dependency on the outside database as part of building the Docker image.  For example, if I try building this branch on my local machine, without the terraso backend container running in the background, the build fails.

josebui:
I think the main disadvantage here is running it on the local environment, I also think this should be executed when the application starts not when the docker image is built, if we can define the command that will be execute it to start the application we can wrap that in a script that will run just for the remote enviroments

paulschreiber:
No longer needed. Render supports a startup command.

PR #435, opened 2023-04-17
fix: project management init migration

david-code:
## Description

I realized that the PR for creating sites had a mistake in the migration; after changing the slug to be a unique ID, I didn't update the migration properly. This PR fixes it, and adds a test to make sure that two sites with the same name can be created.


terraso_backend/apps/project_management/migrations/0001_initial.py
```diff
@@ -46,13 +46,4 @@ class Migration(migrations.Migration):
                 violation_error_message="slug",
             ),
         ),
-        migrations.AddConstraint(
-            model_name="site",
```
josebui, line 50: Question: Do we need a migration to remove this constraint? since it was already on main?

david-code: Yeah, this should actually be dealt with in the next PR I think. Going to close this PR.

PR #437, opened 2023-04-18
feat: add sitemap with landscape profiles

paulschreiber:
## Description
Add a sitemap containing landscape profiles.


terraso_backend/apps/core/models/landscapes.py
```diff
@@ -150,6 +150,10 @@ def get_default_group(self):
 
         return landscape_group.group
 
+    # used for sitemap generation
+    def get_absolute_url(self):
```
josebui, line 154: Question: Where is this function being used? I don't see this referenced in this PR

terraso_backend/apps/core/sites.py
```diff
@@ -0,0 +1,36 @@
...
+# along with this program. If not, see https://www.gnu.org/licenses/.
+
+from django.conf import settings
+from django.contrib.sitemaps import GenericSitemap
+from django.contrib.sites.models import Site
+
+from apps.core.models import Landscape
+
+
+class TerrasoSitemap(GenericSitemap):
```
josebui, line 23: Issue: This is the Web Client Sitemap, it should me named with that in mind, also this should not be part of the core app since this should be explicitly in a place just related to the webclient, also a comment should be added that ideally this code should not be here since the API should be agnostic of the existance of a web client

josebui (changes requested):

terraso_backend/config/urls.py
```diff
@@ -77,4 +79,10 @@
     path("storage/", include("apps.storage.urls", namespace="terraso_storage")),
     path("shared-data/", include("apps.shared_data.urls", namespace="shared_data")),
     path("story-map/", include("apps.story_map.urls", namespace="story_map")),
+    path(
+        "sitemap.xml",
```
josebui, line 83: Issue: This path should also explicitly state that this is related to the web client, also depending on how the rest of the code gets restrucutred this path might not be referenced directly here

josebui (changes requested):

terraso_backend/apps/core/models/landscapes.py
paulschreiber: It is used by the built-in [Django sitemap component](https://docs.djangoproject.com/en/4.2/ref/contrib/sitemaps/). It is called over the iterable (Landscape.objects.all()).

PR #460, opened 2023-04-28
feat: E2E tests backend support

josebui:
## Description

- Added Admin Graph QL schema for admin operations
- Added is_test_user flag to identify test users
- Added function to generate tokens for test users

### Checklist
- [ ] New tests added

### Related Issues
Fixes https://github.com/techmatters/terraso-web-client/issues/819


terraso_backend/apps/core/models/users.py
```diff
@@ -71,6 +71,11 @@ class User(SafeDeleteModel, AbstractUser):
     email = models.EmailField()
     profile_image = models.URLField(blank=True, default="")
 
+    # Alternatives to is_test_user?:
+    # 1. Add a fixed prefix to the email address
+    # 2. Create a separate table to store what users are test users
+    is_test_user = models.BooleanField(blank=True, default=False)
```
paulschreiber, line 77: I don't think it makes sense to add a flag to the model. Having another table might be okay. Please do further research in this area.

terraso_backend/apps/graphql/schema/commons.py
```diff
@@ -62,6 +62,24 @@ def mutate(cls, root, info, input):
...
 
+class BaseAdminMutation(BaseMutation):
+    class Meta:
+        abstract = True
+
+    @classmethod
+    def mutate(cls, root, info, input):
+        user = info.context.user
+
+        if not user or not user.is_authenticated or not user.is_superuser:
```
paulschreiber, line 73: question: what do we need the superuser check for?

josebui: @paulschreiber I think so, for the current available operations added as part of this PR it makes sense to just let superuser access them

terraso_backend/apps/core/models/users.py
josebui: @paulschreiber There is not a lot of specific information about how to implement this, there is a lot of mentions related to isolating test data, so knowing what are the test users and even potentially identify what have they done in the system is important. With that in mind I do agree that the core app should not know details about testing, so I think a separate table makes sense, should this be a separate app intended for e2e tests?

josebui: This is the most relevant conversation I found related to the topic, but no mention about implementation details: https://stackoverflow.com/questions/55093017/should-e2e-tests-persist-data-in-real-databases

david-code (commented):
Hi Jose, overall this looks good. I have a question about an alternative implementation that seems simpler to me, however. Could you instead just:
- create a test user in the database using the `admin.py` script. This user would have a password because of the way the script works.
- have an endpoint (doesn't need to be graphql) that allows the user to authenticate with their password
- the password can be checked using the regular django means and a token generated

I can see that perhaps we don't want to expose a general password authentication endpoint. But are there other drawbacks that would disqualify this approach? I'm curious what you think.

josebui:
@david-code I think this implementation is really similar, in terms of complexity I think they are almost the same, so, from the implemented process to generate a test user token you have to:
1. Login to the admin panel
2. Create a new user with the test user flag enabled
2. Go to `/graphql/admin`
3. Execute the generate token mutation

I think this approch has this advantages:

- We have a graph QL admin schema that we can extend to add other admin processes, I think this is better than having a regular REST endpoint because of the regular Graph QL benefits, also it is easier to user from the graphical interface.
- Creating a login endpoint that is just used from the tests seems to be adding a different feature that is not relevant for real users even if it is kind of available from Django, I think an improvement to this could be implementing the mechanism to use sso to login from the e2e tests, but I think for now that is high effort, and I think of this as a future improvement for the e2e tests.

So, maybe to talk about more about the simplicity side, I think the current implementation is simpler because we don't have to implement the process to login from the e2e tests using this mechanism that is not used by our regular user.

@david-code let me know what do you think

david-code:
Ok, I see. That makes sense. The solution I proposed would would work a bit differently:
- e2e tests take as config the test user username and password
- on setup, test authenticates to get a new user token
- test runs

If I understand correctly, the idea here is more to generate a token that could be used as like an environment variable for the end-to-end tests, is that right? If so, that would work too, but I think a solution that generates a new token for each test is more robust.
(Let me know if I'm misunderstanding with this PR accomplishes).

Either way, I think both solutions would work, and I'd be OK with this one too.

josebui:
@david-code Yes, I agree with that, a next improvement for this will be to figure out how to generate a token for the tests using the sso process. And yes, right now that token will be saved in a secret enviroment variable to be used by the e2e tests

paulschreiber (approved):

shrouxm (commented):
i am missing any context/discussion that went into what our E2E tests will look like, but:

question: are the tests going to run against prod or just in test environments? if just in test environments, do we need to create specific test users or do anything specific to handle them or could we just create normal users?

josebui:
> i am missing any context/discussion that went into what our E2E tests will look like, but:
question: are the tests going to run against prod or just in test environments? if just in test environments, do we need to create specific test users or do anything specific to handle them or could we just create normal users?

@shrouxm  You didn't miss context or a discussion, this is the first step for our e2e tests, so any topic is still open for discussion, and can defenitely be part of future improvements if we need to.

> are the tests going to run against prod or just in test environments?

My expectation is that at this first iteration the tests will not be executed in a production environment just `staging` and/or `testing`

> do we need to create specific test users or do anything specific to handle them or could we just create normal users?

For this implementation the expectation is that we will create users manually, flag them as test users and generate a token to be used by the e2e tests

Let me know if you have any concern about those expectations for this first iteration

shrouxm:
> For this implementation the expectation is that we will create users manually, flag them as test users and generate a token to be used by the e2e tests

that makes sense, it's not a blocking question because i don't think it's a big deal either way and we can just change it later. i guess what i'm wondering is whether the e2e tests could sign in normally to keep the user model simpler, or is that really complicated because OAuth & maybe something we could revisit if we ever implement email + password based sign-in?

PR #479, opened 2023-05-03
chore: add command to generate graphql schema file

shrouxm:
## Description
Questions: 
- `docs.html` and `schema.graphql` are both generated files. `docs.html` was out of date prior to this PR. Should we either gitignore them or add a lint check to ensure they're up to date? 
- The utility of generating the `schema.graphql` is that it can be used for type generation/checking in clients. Should the file be checked into the client?


david-code (approved):
I'm approving but @josebui maybe you would like to take a look as well.

shrouxm:
Ok actually apparently the docs file is used for this page: https://api.terraso.org/graphql/docs. So I will add a lint for both of them, we can revisit if it's annoying.

shrouxm:
Ok so the doc generation is non-deterministic :facepalm: I will not add a lint for it and will file an issue to generate it when building the app which I think is the right way to do it anyway. I am trying to add a lint to check the schema, it works locally but not in CI so this should be ready to merge once I figure that out but I'm going to take a break because this has taken up way too much time already.

PR #481, opened 2023-05-04
types: refine graphql schema connection types

shrouxm:
## Description
Changes the type of our schema's Connection and Edge types from this:
```graphql
type LandscapeGroupNodeConnection {
  """Pagination data for this connection."""
  pageInfo: PageInfo!

  """Contains the nodes in this connection."""
  edges: [LandscapeGroupNodeEdge]!
  totalCount: Int
}

"""A Relay edge containing a `LandscapeGroupNode` and its cursor."""
type LandscapeGroupNodeEdge {
  """The item at the end of the edge"""
  node: LandscapeGroupNode

  """A cursor for use in pagination"""
  cursor: String!
}
```
to this:
```graphql
type LandscapeGroupNodeConnection {
  """Pagination data for this connection."""
  pageInfo: PageInfo!

  """Contains the nodes in this connection."""
  edges: [LandscapeGroupNodeEdge!]!
  totalCount: Int
}

"""A Relay edge containing a `LandscapeGroupNode` and its cursor."""
type LandscapeGroupNodeEdge {
  """The item at the end of the edge"""
  node: LandscapeGroupNode!

  """A cursor for use in pagination"""
  cursor: String!
}
```
(note the extra `!`s on the `edges` and `node` fields). Our frontend code was already assuming this is the case and it is indeed always the case by graphene-django's design. See this issue on graphene-django: https://github.com/graphql-python/graphene-django/issues/901

### Verification steps
All behavior should remain the same. Updated types should be reflected in docs.

### DIFF
--- a/terraso_backend/apps/graphql/schema/commons.py
+++ b/terraso_backend/apps/graphql/schema/commons.py
@@ -43,7 +43,7 @@ class TerrasoConnection(Connection):
     class Meta:
         abstract = True
 
-    total_count = Int()
+    total_count = Int(required=True)
 
     def resolve_total_count(self, info, **kwargs):
         queryset = self.iterable
--- a/terraso_backend/apps/graphql/schema/commons.py
+++ b/terraso_backend/apps/graphql/schema/commons.py
@@ -19,7 +19,7 @@ import re
 import structlog
 from django.core.exceptions import NON_FIELD_ERRORS, ValidationError
 from django.db import IntegrityError
-from graphene import Connection, Int, relay, NonNull, Field, String, List, ObjectType
+from graphene import Connection, Field, Int, List, NonNull, ObjectType, String, relay
 from graphene.types.generic import GenericScalar
 
 from apps.core.formatters import from_camel_to_snake_case
--- a/terraso_backend/apps/graphql/schema/commons.py
+++ b/terraso_backend/apps/graphql/schema/commons.py
@@ -51,13 +51,13 @@ class TerrasoConnection(Connection):
 
     # This will coax graphene to output more precise types for connections.
     # Context: https://github.com/graphql-python/graphene/pull/1504
-    # Will be unnecessary if this lands: https://github.com/graphql-python/graphene-django/issues/901
+    # Will be unnecessary after https://github.com/graphql-python/graphene-django/issues/901
     @classmethod
     def __init_subclass_with_meta__(cls, node=None, **options):
         type_name = re.sub("Connection$", "", cls.__name__)
 
         node_for_edge = node
-        if node != None and not isinstance(node, NonNull):
+        if node is not None and not isinstance(node, NonNull):
             node_for_edge = NonNull(node)
 
         class Edge(ObjectType):
@@ -72,7 +72,7 @@ class TerrasoConnection(Connection):
         cls.Edge = edge_type
 
         cls.edges = Field(
-            NonNull(List(NonNull(edge_type))), 
+            NonNull(List(NonNull(edge_type))),
             description="Contains the nodes in this connection.",
         )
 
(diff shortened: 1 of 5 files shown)


terraso_backend/apps/graphql/schema/commons.py
```diff
@@ -43,6 +44,35 @@ def resolve_total_count(self, info, **kwargs):
...
+        edge_type = type(f"{type_name}Edge", (Edge,), {"Meta": Meta})
+
+        cls.Edge = edge_type
+
+        cls.edges = Field(
+            NonNull(List(NonNull(edge_type))),
+            description="Contains the nodes in this connection.",
+        )
+
+        super(TerrasoConnection, cls).__init_subclass_with_meta__(node=node, **options)
```
david-code, line 74: suggestion: I *think* it is possible to just use [super](https://docs.python.org/3/library/functions.html#super) here, which I find a bit less confusing.

terraso_backend/apps/graphql/schema/commons.py
```diff
@@ -43,6 +44,35 @@ def resolve_total_count(self, info, **kwargs):
         queryset = self.iterable
         return queryset.count()
 
+    # This will coax graphene to output more precise types for connections.
+    # Context: https://github.com/graphql-python/graphene/pull/1504
+    # Will be unnecessary after https://github.com/graphql-python/graphene-django/issues/901
+    @classmethod
+    def __init_subclass_with_meta__(cls, node=None, **options):
+        type_name = re.sub("Connection$", "", cls.__name__)
```
david-code, line 52: suggestion: Maybe it would be a good idea to assert the class ends in `Connection`, just in case.

david-code (approved):
Nice, I'm glad you were able to find a workaround here

terraso_backend/apps/graphql/schema/commons.py
```diff
@@ -37,19 +38,54 @@ class TerrasoConnection(Connection):
     class Meta:
         abstract = True
 
-    total_count = Int()
+    total_count = Int(required=True)
```
josebui, line 41: Question: What is the mening of required here? is this an input value?

shrouxm: `required` is shorthand for wrapping in `NonNull`: https://docs.graphene-python.org/en/latest/types/list-and-nonnull/#nonnull

it's an assertion that the field must resolve to a non-null value, or else the server will raise an error and the request will fail. in this case because it's resolving to a count of an iterable, it will always be true (unless the entire request fails, in which case some enclosing node of response tree will be null). making the assertion here: a) catches potential errors if we accidentally return null on the server, and b) simplifies client code since it no longer needs to handle potential null values on this field

shrouxm:
@josebui `edges: []` conforms to the type `[LandscapeGroupNodeEdge!]!`, since all of the edges in the list are non-null. the update ensures that the list itself is never null, and it never contains nulls, so `edges: null` and `edges: [null]` are not allowed. my understanding is graphene never would've returned either of those (see discussion here: https://github.com/graphql-python/graphene-django/issues/901), but the server will throw an error if that does happen so any bugs it produces won't be subtle

josebui:
Thanks @shrouxm that makes sense

terraso_backend/apps/graphql/schema/commons.py
josebui: @shrouxm do you know if this validation executed on top of the requested fields in the query? or always before resolving the node fields?

terraso_backend/apps/graphql/schema/commons.py
shrouxm: it still works if it doesn't, this code provides a cleaner name if it does (and with how our code is structured it always does)

terraso_backend/apps/graphql/schema/commons.py
shrouxm: @josebui hmm i'm not sure what you mean, like does the validation happen at the very end before returning the result or does it happen right after the resolve function is called? in either case i'm not entirely sure, but maybe i can intentionally return `null` in the resolver and see what happens for our edification

PR #491, opened 2023-05-08
fix: Continue SSO signin provision for users not logged into Terraso

### LINKED ISSUE #478: User not redirected properly after SSO signin if not signed into Terraso before SSO process
## Description

When a site uses Terraso as an SSO provider, the user will be prompted to log into Terraso if not already logged in. The expected response is for the SSO process to continue after logging in with their Terraso identity, be it to grant their permission for Kobo to use their information, or to continue to KoBo itself. The current behaviour is that the user is instead redirected to the Terraso homepage. They have to repeat the process to log into KoBo.

## Steps To Reproduce

1. Log out of your account on Terraso staging (or open a private browser window)
2. Go to test KoBo URL (posted in Slack; not sure if KoBo wants anyone on the Internet to be using it)
3. Initiate Terraso SSO login
4. Observe that you are redirected to the Terraso home page

## Expected behavior
Redirected to KoBo or asked to authorize Kobo to access info

## Actual behavior
Redirected to Terraso main page

## Additional context

This seems quite a bit more complicated than I anticipated. 

An OAuth client provides a `redirect_uri` to the provider. So here Kobo provides a redirect URI where the user will be redirected after a successful authentication step to continue the OAuth process (there's a step asking for access tokens after the authentication), and Terraso passes on its own redirect URI to say Google. Once the user logs into Google, they will be redirected to the Terraso URI. We want Terraso to then redirect the user to the original Kobo redirect URI. We also need to include the relevant OAuth params (`code` and `state`, see [4.1.2 Authorization Response](https://www.rfc-editor.org/rfc/rfc6749#section-4.1.2)).

The problem here is that information in OAuth is passed on by HTTP URL params, body params and headers, and not stored in cookies or in state somewhere. Google OAuth won't send us back extra request body params or headers that we send, and in fact will likely throw an error. That means that we can't store the `redirect_uri` in the requests we are sending to Google.

How can this be done?

I think the cleanest solution would be to store the `redirect_uri` from an OAuth client in a cookie. After a successful Terraso login, we could query to see if this cookie is present and continue the OAuth process with the SSO provider if so. 

We can use [cookie security features](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#cookies) to limit cookie accessibility and prevent it being seen by other websites, set a short expiry date, etc. Even if the cookie were to be accessed by a bad actor, the URI itself would not be useful without the access code (`code` generated by us) and the client secret (held by the OAuth client app). Still, it would require some careful thinking to make sure that this is done securely, as it's not part of the OAuth spec and would need to be implemented by us.

david-code:
## Description

As described in #478, users currently are not redirected to continue the SSO signin process if they are not signed into their Terraso account. This PR aims to fix this by storing the URI used to start the signin in a cookie. When the user has completed logging into Terraso, we then redirect them to this original URI, starting the process.

How it works:
- User tried to query `/oauth/authorize`; because they are not logged in, the backend redirects them to our "login page", which is just `/`
- I added middleware that just checks if a not logged-in user is trying to query the URI that starts the Terraso SSO provider process (`/oauth/authorize`); if so, the URI is stored in a signed cookie. An expiry date of five minutes is set so that the cookie doesn't hand around forever.
-  User completes their login to Terraso
- At the final step of their login to Terraso, we check if this cookie is currently set. If it is, we grab the value of the cookie and make sure it matches what we expect. The redirect URI send to the user is set to the cookie value.
- The cookie is deleted.
- User is redirected to original URI and the SSO provider process starts.

Overall, this process is quite simple, but not without its drawbacks. However, I think this is a good first solution. We can continue thinking of ways to improve this process.

### Checklist
- [x] New tests added

### Related Issues
Fixes #478

### Verification steps
- This currently works with our staging URL and the Kobo test server
- Check slack or ask me for the URL to the test Kobo server
- Open a private window and paste the URL
- Go through the SSO signin process
- You will end up being redirected to Kobo signup screen

### DIFF
--- a/terraso_backend/apps/auth/constants.py
+++ b/terraso_backend/apps/auth/constants.py
@@ -1,0 +1,1 @@
+OAUTH_COOKIE_MAX_AGE_SECONDS = 300  # 5 minutes
--- a/terraso_backend/apps/auth/constants.py
+++ b/terraso_backend/apps/auth/constants.py
@@ -1,1 +1,2 @@
+OAUTH_COOKIE_NAME = "oauth"
 OAUTH_COOKIE_MAX_AGE_SECONDS = 300  # 5 minutes
--- a/terraso_backend/apps/auth/middleware.py
+++ b/terraso_backend/apps/auth/middleware.py
@@ -24,7 +24,7 @@ from django.http.response import JsonResponse
 from django.urls import reverse
 from jwt.exceptions import InvalidTokenError
 
-from .constants import OAUTH_COOKIE_MAX_AGE_SECONDS
+from .constants import OAUTH_COOKIE_MAX_AGE_SECONDS, OAUTH_COOKIE_NAME
 from .services import JWTService
 
 logger = structlog.get_logger(__name__)
@@ -123,19 +123,19 @@ class OAuthAuthorizeState:
 
         if request.path == self.uri_path and request.user.is_anonymous:
             # user accessing OAuth authorize URI and not logged in
-            # they will directed to log in by the frontend
-            # in the mean time, we store the URL so OAuth can start after
+            # we store the URL so OAuth can start after
             # user logged in
             cookie = request.get_full_path_info()
 
             response.set_signed_cookie(
-                "oauth",
+                OAUTH_COOKIE_NAME,
                 cookie,
                 domain=settings.AUTH_COOKIE_DOMAIN,
                 max_age=OAUTH_COOKIE_MAX_AGE_SECONDS,
                 httponly=True,
                 secure=True,
-                # need lax for our usage here
+                # lax - cookie sent from requests not originating from our domain
+                # need this for the oauth flow b/c request coming from third party
                 samesite="Lax",
             )
 
--- a/terraso_backend/apps/auth/middleware.py
+++ b/terraso_backend/apps/auth/middleware.py
@@ -21,8 +21,10 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.models import AnonymousUser
 from django.core.exceptions import ImproperlyConfigured, ValidationError
 from django.http.response import JsonResponse
+from django.urls import reverse
 from jwt.exceptions import InvalidTokenError
 
+from .constants import OAUTH_COOKIE_MAX_AGE_SECONDS
 from .services import JWTService
 
 logger = structlog.get_logger(__name__)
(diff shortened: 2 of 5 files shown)


terraso_backend/apps/auth/middleware.py
```diff
@@ -109,3 +111,32 @@ def wrapped_view(request, *args, **kwargs):
...
+            cookie = request.get_full_path_info()
+
+            response.set_signed_cookie(
+                "oauth",
+                cookie,
+                domain=settings.AUTH_COOKIE_DOMAIN,
+                max_age=OAUTH_COOKIE_MAX_AGE_SECONDS,
+                httponly=True,
+                secure=True,
+                # need lax for our usage here
```
paulschreiber, line 138: issue: Explain why `lax` is needed.

terraso_backend/apps/auth/middleware.py
```diff
@@ -109,3 +111,32 @@ def wrapped_view(request, *args, **kwargs):
...
+        self.uri_path = reverse("oauth2_provider:authorize")
+
+    def __call__(self, request):
+        response = self.get_response(request)
+
+        if request.path == self.uri_path and request.user.is_anonymous:
+            # user accessing OAuth authorize URI and not logged in
+            # they will directed to log in by the frontend
+            # in the mean time, we store the URL so OAuth can start after
+            # user logged in
```
paulschreiber, line 128: issue: Please rewrite this. It is not clear.

terraso_backend/apps/auth/constants.py
```diff
@@ -0,0 +1 @@
+OAUTH_COOKIE_MAX_AGE_SECONDS = 300  # 5 minutes
```
josebui, line 1: Question: What will be affeted by this expiration value?

terraso_backend/apps/auth/views.py
```diff
@@ -92,7 +106,14 @@
...
             return HttpResponse(f"Error: {exc}", status=400)
 
-        response = HttpResponseRedirect(f"{settings.WEB_CLIENT_URL}/{self.state}")
+        redirect_uri = f"{settings.WEB_CLIENT_URL}/{self.state}"
+        if cookie := req.get_signed_cookie("oauth", None, max_age=OAUTH_COOKIE_MAX_AGE_SECONDS):
+            # we stored the original URI as a cookie
+            # redirecting the user to this URI will start our OAuth process
+            if self._check_cookie(cookie):
+                redirect_uri = cookie
+        response = HttpResponseRedirect(redirect_uri)
```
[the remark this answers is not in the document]
josebui: @david-code where is the check of the exepcted url?

terraso_backend/apps/auth/constants.py
paulschreiber: The parameter`max_age=OAUTH_COOKIE_MAX_AGE_SECONDS` in `middleware.py`.

terraso_backend/apps/auth/views.py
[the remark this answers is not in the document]
david-code: The function `_check_cookie` checks to make sure the URL will redirect where we expect it.

terraso_backend/apps/auth/middleware.py
```diff
@@ -109,3 +111,32 @@ def wrapped_view(request, *args, **kwargs):
...
+        self.get_response = get_response
+        self.uri_path = reverse("oauth2_provider:authorize")
+
+    def __call__(self, request):
+        response = self.get_response(request)
+
+        if request.path == self.uri_path and request.user.is_anonymous:
+            # user accessing OAuth authorize URI and not logged in
+            # we store the URL so OAuth can start after
+            # user logged in
```
paulschreiber, line 127:
```suggestion
            # we store the URL so OAuth can start after login
```

paulschreiber (approved):

PR #494, opened 2023-05-09
poc: Generate thumbnail with playwright

josebui:
## Description

- POC to generate a thumbnail using playwright
- Fro the production implementation I expect this to go to a lambda or a fargate process that we can call on demand to generate the thumbnails

### DIFF
--- a/scripts/thumbnail-generator/generate-thumbnail.js
+++ b/scripts/thumbnail-generator/generate-thumbnail.js
@@ -1,0 +1,55 @@
+const { chromium } = require('playwright');
+const { expect } = require('@playwright/test');
+const sharp = require('sharp');
+const fs = require('fs');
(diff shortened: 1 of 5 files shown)


scripts/thumbnail-generator/generate-thumbnail.js
```diff
@@ -0,0 +1,55 @@
...
+  const browser = await chromium.launch();
+  const context = await browser.newContext({
+    viewport: {
+      width: 1280, // Set the desired window width here
+      height: 720, // Set the desired window height here
+    },
+  });
+  const page = await context.newPage();
+
+  await context.addCookies([
```
josebui, line 17: Potentially we could inpersonate a user to get a thumbnail of their private data, it seems that this is not needed for the first iteration

PR #507, opened 2023-05-18
chore: add package.json to allow importing graphql schema

shrouxm:
## Description
This minimal `package.json` will allow other node-based repos to include a reference to the generated schema.graphql file via git reference, rather than needing to manually copy the file over.


package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "terraso-backend",
+  "version": "0.1.0",
+  "private": true,
+  "files": [
+    "terraso_backend/apps/graphql/schema/schema.graphql"
+  ]
+}
```
paulschreiber, line 8:
```suggestion
}

```

paulschreiber (approved):
issue: please add

Issue #402, opened 2023-04-06
GraphQL Mutation to add Site to Project


david-code:
# Spec

 2.6

> As a user with permission to edit the site, I can set the site to be part of any project that I can edit. If I do not own the project, but choose to add my site to the project, I should be warned that I will lose ownership of the site after adding it to the project.

# Definition of Done
- [ ] Frontend can make GraphQL mutation to add Site to Project
- [ ] Only users with edit permissions in a given Project can add the Site to the Project
- [ ] Only users with edit permissions of a given Site can add that Site to another Project

PR #511, opened 2023-05-19
feature: Modify sites model to allow personal sites to be created

### LINKED ISSUE #455: Personal sites can be created
# Description

Sites that a user creates and that are not added to a named project will be treated as personal sites. From there, the user can manipulate them as a manager could if a site were added to a project.

# Design
db_schema

Originally I planned to make the personal sites belong to a personal project. This project would be private, and the settings could not be edited. However, thinking a bit further about this, it is a bit clunky. For example, we need to enforce a constraint on the groups related to the personal projects only having one member, not being deleted, etc. Also the project settings will not make any sense for a private project.

As an alternative, I think it makes sense to associate sites not to a project, but to a site collection (name not decided yet). The idea would be to use the Django ORM inheritance options to have two different models, PersonalSites and ProjectSites that inherit from the model. In SQL, this would mean two different nullable foreign keys with a constraint that one must be not null, I think. ProjectSites would link to a project, while PersonalSites would link to a user. This makes it easier to enforce some constraints (only a single user, no project settings, etc.) with the caveat that it will make checking permissions slightly more clumsy.

# Defintion of Done

- [ ] Other users cannot be added to gain access of these personal sites
- [ ] User can edit and update personal sites

david-code:
## Description

- Modify sites so that they are either owned by a project or a user. This makes personal sites (sites that don't belong to any project) possible
- Make creation sites possible in a given project

### Checklist
- [x] New tests added

### Related Issues
Fixes #455 #402


terraso_backend/tests/graphql/mutations/test_sites.py
```diff
@@ -47,9 +52,21 @@ def test_site_creation(client_query, user):
...
+    response = client_query(CREATE_SITE_QUERY, variables={"input": kwargs})
+    content = json.loads(response.content)
+    assert "errors" not in content and "errors" not in content["data"]
+    id = content["data"]["addSite"]["site"]["id"]
+    site = Site.objects.get(pk=id)
+    assert site.project == project
 
 
-ADD_CLIENT_QUERY = """
+EDIT_SITE_QUERY = """
```
david-code, line 69: This was just incorrectly named. I changed the variable name so it reflects what the mutation actually does (edits the site).

terraso_backend/locale/en/LC_MESSAGES/django.po
```diff
@@ -32,43 +32,59 @@ msgstr ""
...
 msgstr "No"
 
-#: terraso_backend/apps/core/models/landscapes.py:76
+#: terraso_backend/apps/core/models/landscapes.py:79
 msgid "partnership_status.in_progress"
 msgstr "In Progress"
 
-#: terraso_backend/apps/core/models/landscapes.py:77
+#: terraso_backend/apps/core/models/landscapes.py:80
 msgid "partnership_status.yes"
```
david-code, line 88: Not sure why landscape translations are being updated, I didn't update them. If it is important to keep these files up to date we should add a check like a schema check to the CI.

JeancarloBarrios (approved):
LGTM, regarding the translation files, maybe just revert them to main for this PR, I think they are causing a conflict

PR #527, opened 2023-05-24
fix: graphql schema output was not consistent

shrouxm:
## Description
This removes unused i18n code from the landscape model that was tripping up the graphql schema lint step in CI.  

### Related Issues
Unblocks #511 

### Verification steps
Everything should work the same, CI should run successfully.


josebui (approved):
This looks good to me

PR #528, opened 2023-05-25
chore: remove no-longer-used partnership status strings

paulschreiber:
## Description
Followup to #527. Removes corresponding strings from the .po files.


shrouxm (approved):
LGTM

Issue #457, opened 2023-04-26
Split the group model into MembershipList and Group


david-code:
## Plan

- [x] https://github.com/techmatters/terraso-backend/issues/790
- [x] https://github.com/techmatters/terraso-client-shared/issues/93
- [x] https://github.com/techmatters/terraso-mobile-client/issues/410

ltseng:
If this task is no longer valid, can we close it out?

david-code:
It is still valid, but we won't do it right away I think.

ltseng:
Jose has done some work for this, so we will need to revise the size estimate and break down into further subissues as necessary @shrouxm

david-code:
@ltseng fyi the work Jose has done will take care of most of this issue. I'm going to update the description to describe the remaining work to do.

ltseng:
Is there an engineering size associated with this? (Don't want to throw you off your work too much otherwise)

david-code:
@ltseng I'm not sure how to best handle this, I assigned engineering to the sub-tasks I listed in the description instead of this issue.

ltseng:
I keep forgetting to check for sizes on the subissues 😅 ignore me!

PR #530, opened 2023-05-25
fix: s/slug/ID for MembershipAddMutationInput

shrouxm:
## Description
Groups were queried by ID but referenced in the MembershipAddMutationInput by slug, this resolves the inconsistency (& will make PRs for issues like https://github.com/techmatters/terraso-web-client/issues/883 or https://github.com/techmatters/terraso-backend/issues/457 cleaner).

### Checklist
- [ ] DO NOT MERGE UNTIL CORRESPONDING FRONTEND PR IS READY


shrouxm:
actually never mind, i will do this once it's time for the group model refactor

PR #546, opened 2023-05-30
fix: don't do extraneous group existence check

shrouxm:
## Description
Removes an extra check for group existence that was causing a bug in prod backend.

### Checklist
- [ ] New tests added

### Related Issues
Fixes https://github.com/techmatters/terraso-product/issues/143

### Verification steps
Should now be able to add shared links to groups and landscapes again.


shrouxm:
i would like to add a unit test for this, but it seems we don't have a way to unit test our GraphQL mutations, so i will file a follow-up issue to address this instead as i think getting the fix out to prod quickly is more important (i manually verified the fix works in my local environment).

josebui (approved):

PR #549, opened 2023-05-31
chore: improve build caching order in Dockerfile

robert-bo-davis:
## Description

This splits run commands and file copies for faster build times using docker build layer cache

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,10 +18,13 @@ RUN apt-get update && \
 
 WORKDIR /app
 
-COPY --chown=terraso:terraso . /app
+COPY --chown=terraso:terraso requirements.txt /app
+COPY --chown=terraso:terraso Makefile /app
 
 USER terraso
 
 RUN pip install --upgrade pip && make install
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -18,10 +18,13 @@ RUN apt-get update && \
...
 
-COPY --chown=terraso:terraso . /app
+COPY --chown=terraso:terraso requirements.txt /app
+COPY --chown=terraso:terraso Makefile /app
 
 USER terraso
 
 RUN pip install --upgrade pip && make install
 
+COPY --chown=terraso:terraso . /app
```
robert-bo-davis, line 28: Every command adds a new layer to the docker build cache. The cache is invalidated if something changes in the layer before it. Only copying the files needed for `RUN pip install --upgrade pip && make install` before running that means that other file changes in the directory won't invalidate the layer cache. So we do the pip install and then copy the rest of the files and then everything AFTER this step has to run on every build.

paulschreiber (approved):

PR #606, opened 2023-06-20
feat: allow mp4 media type uploads to story maps

jodiezhuu:
## Description
Add a media validation check to handle_config_media to 

### Related PRs
https://github.com/techmatters/terraso-web-client/compare/main...feat/allow-mp4-uploads-to-storymaps

### Related Issues
https://github.com/techmatters/terraso-product/issues/145

### Verification steps
1. On homepage, click on "Create New Story Map"
2. Add a title and click on "Add new chapter"
3. Click on "Add Media" and upload a mp4 file
4. Click "Publish" and should be successful


terraso_backend/apps/story_map/views.py
```diff
@@ -119,12 +119,17 @@ def post(self, request, **kwargs):
...
+        return True
+    logger.info("Invalid media type.")
+    raise Exception("Invalid media type")
 
 def handle_config_media(new_config, current_config, request):
     if "chapters" in new_config:
         for chapter in new_config["chapters"]:
             media = chapter.get("media")
-            if media and "contentId" in media:
+            if media and is_valid_media(media) and "contentId" in media:
```
josebui, line 132: Issue: I think we should do this before the `handle_config_media` function gets called, right now if there are valid media before the system will upload them for S3 and then we will show the user that save failed, I think our options are:
- Omit media files that are not supported without raising an exception, maybe just logging a warning
- Validate before `handle_config_media` if there are any invalid media types and reject the request to save the story map

josebui (changes requested):

terraso_backend/apps/story_map/views.py
```diff
@@ -52,13 +52,28 @@ def post(self, request, **kwargs):
...
 
         entry_form = StoryMapForm(data=form_data)
-
         if not entry_form.is_valid():
             error_messages = get_error_messages(entry_form.errors.as_data())
             return JsonResponse(
                 {"errors": [{"message": [asdict(e) for e in error_messages]}]}, status=400
             )
 
+        if "chapters" in config:
```
josebui, line 61: Issue: Can you create a separate function to hold this code? I also think you need to run this for `StoryMapUpdateView`

terraso_backend/apps/story_map/views.py
```diff
@@ -52,13 +52,28 @@ def post(self, request, **kwargs):
...
+                         or media["type"].startswith("audio")
+                         or media["type"].startswith("video"))
+                ):
+                    logger.info("Warning: invalid media type")
+                    error_message = ErrorMessage(
+                        code="Invalid Media Type",
+                        context=ErrorContext(model="StoryMap", field=NON_FIELD_ERRORS)
+                    )
+                    return JsonResponse({"errors": [{
+                        "message": [asdict(error_message)]}]}, status=400)
```
josebui, line 76: Question: Did you check how this gets to the client, I think ideally the code should be something that we can understand form the code side, because it should be localized in the client side

josebui (changes requested):

terraso_backend/apps/story_map/views.py
jodiezhuu: Invalid media type upload request will be rejected on the backend and a warning will be logged to console

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -72,7 +72,9 @@ def resolve_configuration(self, info):
             if (
                 media
                 and "url" in media
-                and (media["type"].startswith("image") or media["type"].startswith("audio"))
+                and (media["type"].startswith("image")
+                     or media["type"].startswith("audio")
+                     or media["type"].startswith("video"))
```
paulschreiber, line 77:
```suggestion
                and media["type"].startswith(("image", "audio", "video"))
```

terraso_backend/apps/story_map/views.py
```diff
@@ -102,6 +110,15 @@ def post(self, request, **kwargs):
...
         new_config = json.loads(form_data["configuration"])
 
+        if invalid_media_type(new_config):
+            logger.info("Warning: invalid media type")
+            error_message = ErrorMessage(
+                code="Invalid Media Type",
+                context=ErrorContext(model="StoryMap", field=NON_FIELD_ERRORS)
+            )
+            return JsonResponse({"errors": [{
+                "message": [asdict(error_message)]}]}, status=400)
```
paulschreiber, line 120: question: why is this called twice?

terraso_backend/apps/story_map/views.py
```diff
@@ -165,6 +182,20 @@ def handle_config_media(new_config, current_config, request):
...
 
+def invalid_media_type(config):
+    if "chapters" in config:
+        for chapter in config["chapters"]:
+            media = chapter.get("media")
+            if not (
+                media
+                    and (media["type"].startswith("image")
+                         or media["type"].startswith("audio")
+                         or media["type"].startswith("video"))
```
paulschreiber, line 193:
```suggestion
                    and media["type"].startswith(("image", "audio", "video"))
```

paulschreiber (changes requested):

terraso_backend/apps/story_map/views.py
jodiezhuu: The first time this is called is validating the media type added is correct, and the second time this is called is for validating if the new media being uploaded in replace of the old one is correct

terraso_backend/apps/story_map/views.py
```diff
@@ -52,13 +52,21 @@ def post(self, request, **kwargs):
...
         entry_form = StoryMapForm(data=form_data)
-
         if not entry_form.is_valid():
             error_messages = get_error_messages(entry_form.errors.as_data())
             return JsonResponse(
                 {"errors": [{"message": [asdict(e) for e in error_messages]}]}, status=400
             )
 
+        if invalid_media_type(config):
+            logger.info("Warning: invalid media type")
```
josebui, line 62: Suggestion: Can this be a warning instead of an info log?

josebui (commented):
Can we add a test for this?

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -72,7 +72,7 @@ def resolve_configuration(self, info):
             if (
                 media
                 and "url" in media
-                and (media["type"].startswith("image") or media["type"].startswith("audio"))
+                and media["type"].startswith(("image", "audio", "video"))
```
paulschreiber, line 75:
```suggestion
                and media["type"].startswith(("image", "audio"))
```

terraso_backend/apps/story_map/views.py
```diff
@@ -87,7 +95,7 @@ def post(self, request, **kwargs):
                 extra={"user_id": user.pk, "story_map_id": str(story_map.id)},
             )
             error_message = ErrorMessage(
-                code="update", context=ErrorContext(model="StoryMap", field=NON_FIELD_ERRORS)
+                code="update", context=ErrorContext(model="StoryMap", field="new_config")
```
josebui, line 98: Issue: The field should be related to the data the user sent, so I think in this case it should be `configuration` instead of `new_config`

paulschreiber:
@jodiezhuu In the future, please use rebase instead of merge to update your branch from `main`.

josebui (approved):
Looks good, just a suggestion to remove a not needed log

PR #617, opened 2023-06-27
feat: Use Mapbox tilesets for shared data visualizations

josebui:
### Checklist
- [x] New tests added

### Related Issues
Fixes https://github.com/techmatters/terraso-web-client/issues/816


terraso_backend/apps/shared_data/migrations/0010_visualizationconfig_mapbox_tileset_id_status.py
```diff
@@ -0,0 +1,43 @@
...
+
+# Generated by Django 4.2.2 on 2023-06-26 23:21
+
+from django.conf import settings
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
```
david-code, line 24: question: Do you know why the migration depends on the user auth model? Glancing over it I can't see any dependency.

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
```diff
@@ -0,0 +1,161 @@
...
+from apps.shared_data.services import data_entry_upload_service
+
+from .models import VisualizationConfig
+
+logger = structlog.get_logger(__name__)
+
+
+class AsyncTaskHandler:
+    def start_task(self, method, args):
+        t = threading.Thread(target=method, args=[*args], daemon=True)
```
david-code, line 19: question: Is there an advantage to setting `daemon=True` here? Looking at the Python [thread docs](https://docs.python.org/3/library/threading.html#thread-objects), it only says that it affects whether the Python program will exit or not.

This is mostly for my own knowledge. I was wondering if maybe you had to set the threads as daemon threads to make something work.

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
```diff
@@ -0,0 +1,161 @@
...
+
+def get_rows_from_file(data_entry):
+    type = data_entry.resource_type
+    if type.startswith("csv"):
+        file = data_entry_upload_service.get_file(data_entry.s3_object_name, "rt")
+        reader = csv.reader(file)
+        return [row for row in reader]
+    elif type.startswith("xls"):
+        file = data_entry_upload_service.get_file(data_entry.s3_object_name, "rb")
+        wb = openpyxl.load_workbook(file)
```
david-code, line 39: question: Would there be any advantage to using the pandas library here? I'm pretty sure we already have it as a dependency, and it seems like it is able to load [excel files](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_excel.html). But maybe the use case here is simple enough that it's overkill.

david-code (approved):
This looks good to me. I mostly left a few questions, but nothing blocking.

terraso_backend/apps/shared_data/migrations/0010_visualizationconfig_mapbox_tileset_id_status.py
josebui: @david-code that is a good question, I used makemigrations just for the shared_data app, maybe that made this different, I will check it

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
josebui: I didn't know there was a library already in the code that we could use to read excel files, I will check it

terraso_backend/tests/graphql/mutations/test_visualization_config_mutations.py
```diff
@@ -22,7 +23,10 @@
 pytestmark = pytest.mark.django_db
 
 
-def test_visualization_config_add(client_query, visualization_configs, data_entries):
+@mock.patch("apps.graphql.schema.visualization_config.start_create_mapbox_tileset_task")
```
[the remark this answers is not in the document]
josebui: This looks interesting, I will check it

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
josebui: @david-code I am not sure if this will have an impact on how our service runs since my expectation is that on a redeploy or a restart of a service render will not be blocked by the program begin able to finish or not. But, I think from the definition of a daemon thread it makes more sense for this case, I don't think we want to make the threads executing the creation of the tilesets a critical part of the process, we could change that in the future but I think for now it makes sense for them to be threads that will not block the program exit, let me know what do you think

terraso_backend/apps/core/gis/mapbox.py
```diff
@@ -0,0 +1,119 @@
...
+        "name": name,
+        "description": description,
+        "attribution": [{"text": "© Terraso", "link": ""}],
+    }
+
+    tileset_id = id
+    response, status_code = _post_tileset(recipe_json, tileset_id)
+
+    if status_code != 200:
+        raise Exception("Post Mapbox tileset error", response)
```
paulschreiber, line 45: suggestion: better describe the error

terraso_backend/apps/core/gis/mapbox.py
```diff
@@ -0,0 +1,119 @@
...
+
+    response, status_code = _publish_tileset(tileset_id)
+
+    if status_code != 200:
+        raise Exception("Publish Mapbox tileset error", response)
```
paulschreiber, line 50: suggestion: better describe the error

terraso_backend/apps/core/gis/mapbox.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def remove_tileset(id):
+    response, status_code = _delete_tileset(id)
+    if status_code != 200:
+        raise Exception("Delete Mapbox tileset error", response)
+
+    response, status_code = _delete_tileset_source(id)
+    if status_code != 204:
+        raise Exception("Delete Mapbox tileset source error", response)
```
paulschreiber, line 62: suggestion: better describe the error

terraso_backend/apps/core/gis/mapbox.py
```diff
@@ -0,0 +1,119 @@
...
+API_URL = settings.MAPBOX_API_URL
+USERNAME = settings.MAPBOX_USERNAME
+TOKEN = settings.MAPBOX_ACCESS_TOKEN
+
+
+def create_tileset(id, geojson, name, description):
+    tileset_source_id = id
+    response, status_code = _post_tileset_source(geojson, tileset_source_id)
+    if status_code != 200:
+        raise Exception("Post Mapbox tileset source error", response)
```
paulschreiber, line 29: suggestion: better describe the error

terraso_backend/apps/core/gis/mapbox.py
```diff
@@ -0,0 +1,119 @@
...
+
+    recipe_json = {
+        "recipe": {
+            "version": 1,
+            "layers": {id: {"source": response["id"], "minzoom": 0, "maxzoom": 14}},
+        },
+        "name": name,
+        "description": description,
+        "attribution": [{"text": "© Terraso", "link": ""}],
```
paulschreiber, line 38:
```suggestion
        "attribution": [{"text": "© 2023 Terraso", "link": ""}],
```

terraso_backend/apps/core/gis/mapbox.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def get_publish_status(id):
+    url = f"{API_URL}/tilesets/v1/{USERNAME}.{id}/jobs?stage=success&access_token={TOKEN}"
+    r = requests.get(url)
+    response = r.json()
+    status_code = r.status_code
+    if status_code != 200:
+        return False
+    return len(response) > 0
```
paulschreiber, line 74:
```suggestion
    response = requests.get(url)
    response_json = response.json()
    status_code = response.status_code
    if status_code != 200:
        return False
    return len(response_json) > 0
```

terraso_backend/apps/core/gis/mapbox.py
```diff
@@ -0,0 +1,119 @@
...
+
+def _post_tileset_source(geojson, id):
+    line_delimited_geojson = "\n".join([json.dumps(feature) for feature in geojson["features"]])
+
+    url = f"{API_URL}/tilesets/v1/sources/{USERNAME}/{id}?access_token={TOKEN}"
+    multipart_data = [("file", ("test.ndjson", line_delimited_geojson, "text/plain"))]
+
+    r = requests.post(url, files=multipart_data)
+    response = r.json()
+    status_code = r.status_code
```
paulschreiber, line 85: suggestion: as above, avoid single-character variable names.

terraso_backend/apps/core/gis/mapbox.py
```diff
@@ -0,0 +1,119 @@
...
+    return response, status_code
+
+
+def _delete_tileset_source(id):
+    url = f"{API_URL}/tilesets/v1/sources/{USERNAME}/{id}?access_token={TOKEN}"
+    r = requests.delete(url)
+    status_code = r.status_code
+    return None, status_code
```
paulschreiber, line 93:
```suggestion
    response = requests.delete(url)
    status_code = response.status_code
    return None, status_code
```

terraso_backend/apps/core/gis/mapbox.py
```diff
@@ -0,0 +1,119 @@
...
+
+def _post_tileset(recipe_json, id):
+    url = f"{API_URL}/tilesets/v1/{USERNAME}.{id}?access_token={TOKEN}"
+    headers = {
+        "Content-Type": "application/json",
+    }
+    r = requests.post(url, json=recipe_json, headers=headers)
+    response = r.json()
+    status_code = r.status_code
+    return response, status_code
```
paulschreiber, line 104: suggestion: as above, avoid single-character variable names.

terraso_backend/apps/core/gis/mapbox.py
```diff
@@ -0,0 +1,119 @@
...
+    status_code = r.status_code
+    return None, status_code
+
+
+def _publish_tileset(id):
+    url = f"{API_URL}/tilesets/v1/{USERNAME}.{id}/publish?access_token={TOKEN}"
+    r = requests.post(url)
+    response = r.json()
+    status_code = r.status_code
+    return response, status_code
```
paulschreiber, line 119: suggestion: as above, avoid single-character variable names.

terraso_backend/apps/core/gis/mapbox.py
```diff
@@ -0,0 +1,119 @@
...
+    r = requests.post(url, json=recipe_json, headers=headers)
+    response = r.json()
+    status_code = r.status_code
+    return response, status_code
+
+
+def _delete_tileset(id):
+    url = f"{API_URL}/tilesets/v1/{USERNAME}.{id}?access_token={TOKEN}"
+    r = requests.delete(url)
+    status_code = r.status_code
```
paulschreiber, line 110:
```suggestion
    response = requests.delete(url)
    status_code = response.status_code
```

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+def get_rows_from_file(data_entry):
+    type = data_entry.resource_type
+    if type.startswith("csv"):
+        file = data_entry_upload_service.get_file(data_entry.s3_object_name, "rt")
+        reader = csv.reader(file)
+        return [row for row in reader]
+    elif type.startswith("xls"):
+        file = data_entry_upload_service.get_file(data_entry.s3_object_name, "rb")
```
paulschreiber, line 38:
```suggestion
    file = data_entry_upload_service.get_file(data_entry.s3_object_name, "rb")
    if type.startswith("csv"):
        reader = csv.reader(file)
        return [row for row in reader]
    elif type.startswith("xls"):
```

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
```diff
@@ -0,0 +1,161 @@
...
+        remove_tileset(tileset_id)
+    except Exception as error:
+        logger.exception(
+            "Error deleting mapbox tileset",
+            extra={"tileset_id": tileset_id, "error": str(error)},
+        )
+
+
+def create_mapbox_tileset(visualization_id):
+    logger.info("Creating mapbox tileset", visualization_id=visualization_id)
```
paulschreiber, line 62: question: does this need to be in production? should it be logger.debug?

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
```diff
@@ -0,0 +1,161 @@
...
+    visualization = VisualizationConfig.objects.get(pk=visualization_id)
+    data_entry = visualization.data_entry
+    group_entry = visualization.group
+
+    # Delete mapbox tileset if needed
```
paulschreiber, line 67: question: what does "if needed" mean?

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
```diff
@@ -0,0 +1,161 @@
...
+
+        geojson = {
+            "type": "FeatureCollection",
+            "features": features,
+        }
+        logger.info(
+            "Geojson generated for mapbox tileset",
+            visualization_id=visualization_id,
+            rows=len(rows),
+        )
```
paulschreiber, line 140: question: does this need to be in production? should it be logger.debug?

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
```diff
@@ -0,0 +1,161 @@
...
+        title = f"{settings.ENV} - {visualization.title}"[:64]
+        description = f"{settings.ENV} - {group_entry.name} - {visualization.title}"
+        id = str(visualization.id).replace("-", "")
+        tileset_id = create_tileset(id, geojson, title, description)
+        logger.info(
```
paulschreiber, line 145: question: does this need to be in production? should it be logger.debug?

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
```diff
@@ -0,0 +1,161 @@
...
+            "Mapbox tileset created",
+            visualization_id=visualization_id,
+            tileset_id=tileset_id,
+        )
+        visualization.mapbox_tileset_id = tileset_id
+        visualization.save()
+        logger.info(
```
paulschreiber, line 152: question: does this need to be in production? should it be logger.debug?

terraso_backend/apps/shared_data/migrations/0010_visualizationconfig_mapbox_tileset_id_status.py
josebui: I removed this, it is not needed

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
josebui: I changed the code to use Pandas

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
josebui: @paulschreiber that refactor is not possible because the read mode is different, one read the file as text `rt` and the other one as binary `rb`

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
josebui: @paulschreiber As right now this is implemented as an async process that the progress will not be displayed to the user, I think it is important to add these logs as info, to potentially check what happened with a tileset, from my experience I have done this not just for this type of async processes but for a lot of the functionality, so this in general feels like a good practice to me, let me know what do you think

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
josebui: I will expand this comment, it means that it will not always delete the tileset, only if there is one to delete

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
josebui: Same answer as before

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
josebui: Same answer as before

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
josebui: Same answer as before

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
paulschreiber: I know you can't read binary files in mode, but I believe you can read text files in binary mode, no?

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
josebui: @paulschreiber it is needed if the tileset is updated, we have to delete it and create it again

paulschreiber: Please include that information in the comment.

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
josebui: Let me try that, not sure

josebui: @paulschreiber No, csv.reader() doesn't accept a binary file

josebui: From the [documentation](https://docs.python.org/3/library/csv.html#csv.reader):

> csvfile can be any object which supports the [iterator](https://docs.python.org/3/glossary.html#term-iterator) protocol and returns a string each time its __next__() method is called

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
```diff
@@ -0,0 +1,162 @@
...
+            "type": "FeatureCollection",
+            "features": features,
+        }
+        logger.info(
+            "Geojson generated for mapbox tileset",
+            visualization_id=visualization_id,
+            rows=len(rows),
+        )
+        title = f"{settings.ENV} - {visualization.title}"[:64]
+        description = f"{settings.ENV} - {group_entry.name} - {visualization.title}"
```
paulschreiber, line 143: question: why is the environment part of the title and description?

josebui: @paulschreiber We just have one mapbox account, so it is important to be able to distinguish from what environment a specific tileset came from

paulschreiber: Tell me more about that. Are the environments billed differently? Or do they have different tiles?

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
josebui: @david-code Hey David let me know what do you think about this

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def create_mapbox_tileset(visualization_id):
+    logger.info("Creating mapbox tileset", visualization_id=visualization_id)
+    visualization = VisualizationConfig.objects.get(pk=visualization_id)
+    data_entry = visualization.data_entry
+    group_entry = visualization.group
+
+    # Delete mapbox tileset if updated tileset, to update a tileset we have
+    # to delete it and create a new one
```
paulschreiber, line 68:
```suggestion
    # You cannot update a Mapbox tileset. We have to delete it and create a new one.
```

terraso_backend/apps/shared_data/visualization_tileset_tasks.py
```diff
@@ -0,0 +1,167 @@
...
+        }
+        logger.info(
+            "Geojson generated for mapbox tileset",
+            visualization_id=visualization_id,
+            rows=len(rows),
+        )
+
+        # Title and description goes to mapbox API, it is important to add
+        # the environment to the title to be able to disntinguish between
+        # different environments in the mapbox studio UI
```
paulschreiber, line 145:
```suggestion
        # Include the environment in the title and description when calling the Mapbox API.
        # Adding the environment to the title allows us to distinguish between environments
        # in the Mapbox studio UI.
```

paulschreiber (approved):

PR #618, opened 2023-06-28
feat: Mobile sign in

david-code:
### Web OAuth

Currently Terraso follows the typical OAuth process. It works basically as follows:
    - User signs in with OAuth Provider on their website
    - User redirected to our redirect URI (something like `/oauth/<provider>/callback` with an authorization grant
    - Authorization grant exchanged for tokens from OAuth Provider
    - At the end of the process, user logged into Terraso by assigning the cookies `rtoken` and `atoken`. An account is created if not existing for the email address.

### Problems with mobile

The problem with using this workflow on mobile is the redirect URI. Because the mobile device is working as the user agent, and not the browser, we can't use https URIs as the redirect URI, but instead have to provide a custom schema. Our app registers on the device to handle this schema through hard linking. For example, it may be that we will handle `org.terraso:/googleoauth`. But because it is the mobile device that is acting as the user agent pushing the OAuth process along, we have no way of executing code on the Terraso server at the end of the OAuth process. 

Why can't we use the browser as the user agent, like we do on web? There are still browsers on mobile, obviously. The problem is that there is no way for an app to claim cookies set in the browser as belonging to the app. So even if we complete the OAuth process in a mobile browser, we have no way of gathering the necessary tokens at the end of the process. People used to get around this by creating a WebView in the app itself and grabbing the tokens from there, but this sidesteps one of the main advantages of OAuth, the user not needing to identify themself with the provider credentials to the client app itself. In other words, the client app could just as easily grab the password the user is entering to the OAuth provider app... Most OAuth providers will not continue the process if they detect the user is not using a proper browser.

In summary, how OAuth works on mobile:
- Mobile app launches a separate web browser where the user authenticated with the OAuth provider. This is closed when the authentication with the OAuth provider is complete.
- Authorization grant forwarded to custom uri schema to be handled by mobile app itself.
- Mobile app exchanges authorization grant for tokens.

### Solution

The simple solution is to exchange the id_token obtained at the end of the OAuth signup with a Terraso access token.

What is an id token? This is something defined as part of OpenID Connect, which most OAuth providers implement nowadays. There is an [ID token spec](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) which outlines the various properties included within. For our uses, there are two important facts:
- It contains info on the user email, name, profile image, as well some other things we might request
- We can use the public keys published by the provider to verify the provenance of the ID token. The OAuth provider signs the JWT with their private key, and we can use the public key to ensure the contents have not been modified.

I got the basics for this approach from a [Google guide on authenticating with backend servers](https://developers.google.com/identity/sign-in/android/backend-auth). It is pretty straightforward to implement, as can be seen in the PR code itself.

### DIFF
--- a/terraso_backend/apps/auth/urls.py
+++ b/terraso_backend/apps/auth/urls.py
@@ -25,6 +25,7 @@ from apps.auth.views import (
     MicrosoftAuthorizeView,
     MicrosoftCallbackView,
     RefreshAccessTokenView,
+    TokenExchangeView,
 )
 
 app_name = "apps.auth"
@@ -54,4 +55,5 @@ urlpatterns = [
     ),
     path("tokens", csrf_exempt(RefreshAccessTokenView.as_view()), name="tokens"),
     path("logout", csrf_exempt(LogoutView.as_view()), name="logout"),
+    path("token-exchange", csrf_exempt(TokenExchangeView.as_view()), name="token-exchange"),
 ]
--- a/terraso_backend/apps/auth/views.py
+++ b/terraso_backend/apps/auth/views.py
@@ -18,6 +18,7 @@ import json
 from urllib.parse import urlparse
 
 import httpx
+import jwt
 import structlog
 from django.conf import settings
 from django.contrib.auth import get_user_model
@@ -243,3 +244,48 @@ def terraso_login(request, user):
     dj_login(request, user, backend="django.contrib.auth.backends.ModelBackend")
 
     return access_token, refresh_token
+
+
+class TokenExchangeView(View):
+    @staticmethod
+    def _check_request(contents):
+        missing_keys = []
+        for key in ("jwt", "provider"):
+            if key not in contents:
+                missing_keys.append(key)
+        if missing_keys:
+            return JsonResponse({"missing_keys": missing_keys}, status=400)
+        provider = contents["provider"]
+        if provider not in settings.JWT_EXCHANGE_PROVIDERS:
+            return JsonResponse({"bad_provider": provider}, status=400)
+
+    @staticmethod
+    def _get_signing_key(token, provider_url):
+        # fetch jwks
+        jwks_client = jwt.PyJWKClient(provider_url)
+        return jwks_client.get_signing_key_from_jwt(token)
+
+    @staticmethod
+    def _verify_payload(token, signing_key, client_id):
+        algorithms = [signing_key.key_type]
+        return jwt.decode(token, signing_key.key, algorithms=algorithms, audience=client_id)
+
+    def post(self, request, *args, **kwargs):
+        contents = json.loads(request.body)
+        if resp := self._check_request(contents):
+            return resp
+        provider_name = contents["provider"]
+        provider = settings.JWT_EXCHANGE_PROVIDERS[provider_name]
+        try:
+            signing_key = self._get_signing_key(contents["jwt"], provider_url=provider["url"])
+        except Exception:
+            return JsonResponse(
+                {"jwk_error": f"could not retrieve signing key for {provider_name}"}, status=500
+            )
+        try:
+            payload = self._verify_payload(contents["jwt"], signing_key, provider["client_id"])
+        except Exception:
+            return JsonResponse({"token_error": "token was not verified"})
+        user = self._create_or_fetch_user(payload)
+        rtoken, atoken = terraso_login(user, request)
+        return JsonResponse({"rtoken": rtoken, "atoken": atoken})
--- a/terraso_backend/apps/auth/views.py
+++ b/terraso_backend/apps/auth/views.py
@@ -15,6 +15,7 @@
 
 import functools
 import json
+from typing import Optional
 from urllib.parse import urlparse
 
 import httpx
@@ -267,8 +268,24 @@ class TokenExchangeView(View):
 
     @staticmethod
     def _verify_payload(token, signing_key, client_id):
-        algorithms = [signing_key.key_type]
+        algorithms = [signing_key._jwk_data.get("alg", "RS256")]
         return jwt.decode(token, signing_key.key, algorithms=algorithms, audience=client_id)
+
+    @staticmethod
+    def _create_or_fetch_user(
+        email: str = "",
+        first_name: str = "",
+        last_name: str = "",
+        picture: Optional[str] = None,
+        **kwargs,
+    ):
+        account_service = AccountService()
+        additional_kwargs = {}
+        if picture:
+            additional_kwargs["profile_image_url"] = picture
+        # TODO: using a private method of AccountService is weird, should be refactored
+        user, created = account_service._persist_user(email, first_name, last_name)
+        return user, created
 
     def post(self, request, *args, **kwargs):
         contents = json.loads(request.body)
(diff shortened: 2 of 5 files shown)


david-code:
@josebui I'm going to take a look later, but would you happen to know why the [setup of test_terms_query](https://github.com/techmatters/terraso-backend/actions/runs/5417310964/jobs/9848185568?pr=618#step:4:374) might be failing? I'm thinking I probably messed up a merge somewhere. I'll take a look later, but I thought I'd ping you just in case the error was obvious to you.

terraso_backend/config/settings.py
```diff
@@ -364,3 +365,17 @@
     "/healthz/",
     "/web-client/sitemap.xml",
 ]
+
+
+class JWTProvider(TypedDict):
+    url: str
+    client_id: str
+    name: str
+
```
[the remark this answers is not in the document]
david-code: I made this is a type hint, so that it's obvious to users what type to use for the mobile settings. I think it might be a good idea to add a comment explaining that.

terraso_backend/apps/auth/services.py
```diff
@@ -250,3 +250,60 @@ def track_signup(self, auth_provider: str, req) -> None:
...
+
+    @staticmethod
+    def _get_signing_key(token, provider_url):
+        # fetch jwks
+        jwks_client = jwt.PyJWKClient(provider_url)
+        return jwks_client.get_signing_key_from_jwt(token)
+
+    @staticmethod
+    def _verify_payload(token, signing_key, client_id):
+        algorithms = [signing_key._jwk_data.get("alg", "RS256")]
```
josebui, line 293: Question: Why is RS256 the default algorithm? it seems that we are using a different default for our current auth process, see: https://github.com/techmatters/terraso-backend/blob/main/terraso_backend/config/settings.py#L283

terraso_backend/apps/auth/services.py
```diff
@@ -250,3 +250,60 @@ def track_signup(self, auth_provider: str, req) -> None:
...
+
+    def validate(self):
+        try:
+            signing_key = self._get_signing_key(self.token, self.jwks_uri)
+        except Exception:
+            msg = f"could not retrieve signing key for {self.provider_name}"
+            logger.exception(msg)
+            raise TokenExchangeException(message=msg, error_type="jwks_error")
+        try:
+            payload = self._verify_payload(self.token, signing_key, self.client_id)
```
josebui, line 304: Suggestion: I think it could be better i we catch the specific expections thrown by JWT, I think you can get them from: from jwt.exceptions

terraso_backend/apps/auth/services.py
```diff
@@ -250,3 +250,60 @@ def track_signup(self, auth_provider: str, req) -> None:
...
+    @staticmethod
+    def _get_signing_key(token, provider_url):
+        # fetch jwks
+        jwks_client = jwt.PyJWKClient(provider_url)
+        return jwks_client.get_signing_key_from_jwt(token)
+
+    @staticmethod
+    def _verify_payload(token, signing_key, client_id):
+        algorithms = [signing_key._jwk_data.get("alg", "RS256")]
+        return jwt.decode(token, signing_key.key, algorithms=algorithms, audience=client_id)
```
josebui, line 294: Question: Our current auth process is using: JWTService().verify_token(token) does it make sense to reuse the same logic to validate JWT tokens?

terraso_backend/apps/auth/urls.py
```diff
@@ -54,4 +55,5 @@
     ),
     path("tokens", csrf_exempt(RefreshAccessTokenView.as_view()), name="tokens"),
     path("logout", csrf_exempt(LogoutView.as_view()), name="logout"),
+    path("token-exchange", csrf_exempt(TokenExchangeView.as_view()), name="token-exchange"),
```
josebui, line 58: Question: I just want to see if I understood this correctly, so, at the end of the OAuth flow the mobile app gets an id_token from the OAuth provider, then with that token the mobile app will create a request to `token-exchange` to get a terraso token, is that correct?

josebui:
@david-code This looks good, I am just not sure about how secure the exchange of the tokens is because I don't have experience with authentication on the mobile app side. So, as far as I can see our protection is the signing of the JWT from the provider and us being able to validate if that id_token was not changed, is that correct? If that is the case is there a chance for another app to impersonate our app and get an id_token (without changing it) and then sending it to our backend to get a terraso toekn? is that an issue? or what are the risks of having that id_token in the mobile app?

josebui:
@david-code about the test failure it seems that a fixture is triggering a unique constraint validation for a user entry, not sure how that can happen

terraso_backend/apps/auth/services.py
david-code: I set this as the default because that is what was using by Google. Looking at this again, it probably makes more sense to throw an exception if the key is not present - it will be easier to read the logs and figure out what is going on.

terraso_backend/apps/auth/services.py
david-code: I think the JWTService uses our own private key to encrypt the JWTs. Here we will use Google's public key to check the JWT's signature and supply Google's client id to ensure that the JWT was created for our own OAuth application, so the use is slightly different.

terraso_backend/apps/auth/urls.py
david-code: Yep, that's correct.

david-code:
> So, as far as I can see our protection is the signing of the JWT from the provider and us being able to validate if that id_token was not changed, is that correct?

Yep, that's right. Someone needs to sign a hash of JWT payload with, for example Google's private key. We can then use Google's public key to check that the JWT signature is correct and the payload hasn't been tampered with.

> If that is the case is there a chance for another app to impersonate our app and get an id_token (without changing it) and then sending it to our backend to get a terraso toekn? is that an issue? or what are the risks of having that id_token in the mobile app?

I do think there are some protections against this. Google, for example, requires you to supply a package ID that your app uses to register itself, and will restrict OAuth signin if the OAuth app package ID doesn't match what was supplied in the OAuth app registration.

However, I think even if a user created their own custom app, somehow tricked Google into thinking their app ID matched the Terraso app ID and got an ID token, this would not be so bad. Afterall, they need to prove to Google that they control the account they are using to authenticated; at that point, they would be able to log into Terraso anyways.

This pattern also is suggested by [Google docs](https://developers.google.com/identity/sign-in/android/backend-auth), and I also have found it being used in a [popular open source React native app](https://github.com/artsy/eigen/blob/435229da25e2f22f08b8236b831e54ace80258ab/src/app/store/AuthModel.ts#L371), which makes me more confident.

terraso_backend/apps/auth/views.py
```diff
@@ -243,3 +250,68 @@ def terraso_login(request, user):
...
+        if missing_keys:
+            return JsonResponse({"missing_keys": missing_keys}, status=400)
+        provider = contents["provider"]
+        if provider not in settings.JWT_EXCHANGE_PROVIDERS:
+            return JsonResponse({"bad_provider": provider}, status=400)
+
+    @staticmethod
+    def _token_error(e):
+        match e.error_type:
+            case "jwks_error" | "bad_config":
```
paulschreiber, line 290: issue: this looks like a bitwise or, which is probably not what you want for strings.

terraso_backend/config/settings.py
```diff
@@ -365,3 +366,19 @@
...
+    url: str
+    client_id: str
+    name: str
+
+
+JWT_EXCHANGE_PROVIDERS: dict[str, JWTProvider] = {
+    "google": dict(
+        url="https://www.googleapis.com/oauth2/v3/certs",
+        client_id=config("GOOGLE_MOBILE_CLIENT_ID", default=""),
```
paulschreiber, line 382: question: Do we need a default? Shouldn't the config file gave a fallback value already?

paulschreiber (changes requested):

terraso_backend/apps/auth/views.py
david-code: This is actually part of [Python pattern matching](https://peps.python.org/pep-0636/#or-patterns)

terraso_backend/config/settings.py
david-code: Yes, but the config values aren't read in unit tests, which causes an exception to be thrown when the env variable doesn't have a value. We could make a separate issue to create an auto-loaded pytest fixture that loaded the env file if we think it would be useful.

paulschreiber (approved):

PR #624, opened 2023-06-30
Feat: invalidate oversized files

### LINKED ISSUE #569: Validate file size for uploaded files
## Description
Right now there is no validation for uploaded files size in the API, we should be able to have a configurable value for the max size allowed. Part of this work should be to indentify all the endpoints that allow file upload.

jodiezhuu:
## Description
Validate file upload size before API call at each upload file endpoint

### Related Issues
Fixes #569


josebui (approved):
Looks good

terraso_backend/apps/storage/file_utils.py
```diff
@@ -0,0 +1,22 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see https://www.gnu.org/licenses/.
+
+def is_file_upload_oversized(files, max_size):
+    for file in files:
+        return get_file_size(file) > max_size
```
paulschreiber, line 18: suggestion: `get_file_size` is unnecessary. Reference `file.size` directly.

terraso_backend/apps/storage/file_utils.py
```diff
@@ -0,0 +1,22 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see https://www.gnu.org/licenses/.
+
+def is_file_upload_oversized(files, max_size):
+    for file in files:
```
paulschreiber, line 17: question: what happens if there is more that one file? what if the second file is oversize but the first one is not? what is the expected behaviour here?

terraso_backend/apps/storage/file_utils.py
jodiezhuu: @paulschreiber We need the get_file_size function to mock an oversized file when unit testing: Check `terraso_backend/tests/shared_data/test_views.py`

PR #639, opened 2023-07-10
feat: Add project management models to Django admin

david-code:
## Description

This was a quick convenience add so we can add models via the Django admin view, which will simplify development now that we will be interacting with the backend more often. Unfortunately there was some other stuff that came up as well:
- Project settings had been accidentally removed as a setting from the Project model. I had to add it back, which takes 3 migrations. Also has to remove a `created_by` field that was removed and slipped back in.
- There seems to be some formatting errors that had slipped into the code, which makes me think that the CI is perhaps not catching them properly. I think this is worth looking into more.


terraso_backend/apps/project_management/models/projects.py
```diff
@@ -62,7 +62,9 @@ def default_settings():
         ProjectSettings, on_delete=models.PROTECT, default=default_settings
     )
 
-    archived = models.BooleanField(default=False,)
+    archived = models.BooleanField(
+        default=False,
+    )
```
paulschreiber, line 67:
```suggestion
    archived = models.BooleanField(default=False)
```

paulschreiber (commented):
There's a lot of formatting noise in here. Is that expected? Should it be in a separate PR?

david-code:
@paulschreiber No, I left a comment about that too. I'm not sure why this is the case, I just ran `make format`. Like I wrote, maybe the linting check is broken in the CI and some commits that aren't properly formatted slipped in.

terraso_backend/apps/project_management/migrations/0011_populate_null_projectsettings.py
```diff
@@ -0,0 +1,37 @@
...
+
+from django.db import migrations
+
+
+def assign_unique_project_settings(apps, schema_editor):
+    Project = apps.get_model("project_management", "Project")
+    ProjectSettings = apps.get_model("project_management", "ProjectSettings")
+    db_alias = schema_editor.connection.alias
+    for project in Project.objects.using(db_alias).all():
+        project.settings = ProjectSettings.objects.using(db_alias).create(project=project)
```
josebui, line 24: Question: Is this a bideractional relationship between project and project settings? how does that work?

josebui (approved):
Looks good, just some questions out of curiosity

terraso_backend/apps/project_management/models/sites.py
```diff
@@ -53,15 +52,6 @@ class Meta(BaseModel.Meta):
         verbose_name="owner to which the site belongs",
     )
 
-    created_by = models.ForeignKey(
```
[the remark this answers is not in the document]
david-code: Instead of using `created_by`, the site can be owned by a person (`owner` field) or a project (`project`). Basically the created_by was changed to owner to be a bit clearer, and it slipped back in by accident. I had to remove it here to fix up some tests.

terraso_backend/apps/project_management/migrations/0011_populate_null_projectsettings.py
david-code: Yeah, I set up the field to use `OneToOneField` so you can have a 1-1 relationship.

PR #660, opened 2023-07-12
fix: Add project description to project GraphQL node

david-code:
Forgot to update the GraphQL Node fields as well


josebui (approved):

shrouxm (approved):
looks good! i guess this would be on the model, but can description be required with default empty string so there aren't two possible empty values for it?

david-code:
> looks good! i guess this would be on the model, but can description be required with default empty string so there aren't two possible empty values for it?

Oh yeah, good point. Let's do that.

PR #661, opened 2023-07-13
feat: add log when a site is transferred

JeancarloBarrios:
### Checklist
- [ ] New tests added

### Related Issues
Fixes #219


david-code (commented):
The logic looks good, but this should be `SiteEditMutation`. For transferring, we want to track if the site is being moved between projects, so we should log every time the SiteEditMutation args include a different project_id (or owner_id).

david-code (approved):

PR #674, opened 2023-07-18
feat: add backend checks for shared data uploads

### LINKED ISSUE #197: Restrict upload file types to only allowed types

jodiezhuu:
## Description
Restrict file uploads to allowed extensions list.

### Related Issues
Fixes #197 and extends from #198

### DIFF
--- a/terraso_backend/apps/shared_data/views.py
+++ b/terraso_backend/apps/shared_data/views.py
@@ -74,12 +74,9 @@ class DataEntryFileUploadView(AuthenticationRequiredMixin, FormView):
 
 
 def is_valid_shared_data_type(files):
-    for file in files:
-        extension = Path(file.content_type)
-        type = "." + extension.parts[-1]
-        if type not in DATA_ENTRY_ACCEPTED_EXTENSIONS:
-            return False
-    return True
+    return all(
+        "." + Path(file.content_type).name in DATA_ENTRY_ACCEPTED_EXTENSIONS for file in files
+    )
 
 
 def get_error_messages(validation_errors):
--- a/terraso_backend/apps/shared_data/views.py
+++ b/terraso_backend/apps/shared_data/views.py
@@ -76,9 +76,7 @@ class DataEntryFileUploadView(AuthenticationRequiredMixin, FormView):
 def is_valid_shared_data_type(files):
     for file in files:
         extension = Path(file.content_type)
-        logger.info("EXTENSION")
         type = "." + extension.parts[-1]
-        logger.info(type)
         if type not in DATA_ENTRY_ACCEPTED_EXTENSIONS:
             return False
     return True
--- a/terraso_backend/apps/shared_data/views.py
+++ b/terraso_backend/apps/shared_data/views.py
@@ -55,7 +55,7 @@ class DataEntryFileUploadView(AuthenticationRequiredMixin, FormView):
             return JsonResponse({"errors": [{"message": [asdict(error_message)]}]}, status=400)
         if not is_valid_shared_data_type(request.FILES.getlist("data_file")):
             error_message = ErrorMessage(
-                code="Invalid media type",
+                code="invalid_media_type",
                 context=ErrorContext(model="Shared Data", field="context_type"),
             )
             return JsonResponse({"errors": [{"message": [asdict(error_message)]}]}, status=400)
--- a/terraso_backend/apps/shared_data/views.py
+++ b/terraso_backend/apps/shared_data/views.py
@@ -74,9 +74,7 @@ class DataEntryFileUploadView(AuthenticationRequiredMixin, FormView):
 
 
 def is_valid_shared_data_type(files):
-    return all(
-        Path(str(file)).suffix in DATA_ENTRY_ACCEPTED_EXTENSIONS for file in files
-    )
+    return all(Path(str(file)).suffix in DATA_ENTRY_ACCEPTED_EXTENSIONS for file in files)
 
 
 def get_error_messages(validation_errors):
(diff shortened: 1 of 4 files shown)


terraso_backend/apps/shared_data/views.py
```diff
@@ -66,6 +72,14 @@ def post(self, request, **kwargs):
         return JsonResponse(data_entry.to_dict(), status=201)
 
 
+def invalid_media_types(files):
+    for file in files:
+        type = "." + file.content_type.split("/")[1]
```
paulschreiber, line 77: issue: use `pathlib`to get the file extension.

terraso_backend/apps/shared_data/views.py
```diff
@@ -66,6 +72,14 @@ def post(self, request, **kwargs):
         return JsonResponse(data_entry.to_dict(), status=201)
 
 
+def invalid_media_types(files):
```
paulschreiber, line 75: issue: please rename this function so it accurately describes purpose. (1) not all of these file types are "media" (2) functions that return a boolean often use the `is_xyz` form. Something like  `is_valid_shared_file_type` would make sense. (Note that you'd have to add a `not` to line 55.)

paulschreiber (changes requested):

terraso_backend/config/settings.py
```diff
@@ -326,6 +326,24 @@
 DATA_ENTRY_ACCEPTED_TYPES = (
     DATA_ENTRY_DOCUMENT_TYPES | DATA_ENTRY_SPREADSHEET_TYPES | DATA_ENTRY_GIS_TYPES
 )
+DATA_ENTRY_ACCEPTED_EXTENSIONS = [
```
paulschreiber, line 329: Let's get this from `DATA_ENTRY_ACCEPTED_TYPES.keys()`

paulschreiber (changes requested):

terraso_backend/apps/shared_data/views.py
```diff
@@ -66,6 +73,15 @@ def post(self, request, **kwargs):
         return JsonResponse(data_entry.to_dict(), status=201)
 
 
+def is_valid_shared_data_type(files):
+    for file in files:
+        extension = Path(file.content_type)
+        type = "." + extension.parts[-1]
+        if type not in DATA_ENTRY_ACCEPTED_EXTENSIONS:
+            return False
+    return True
```
josebui, line 82: Suggestion: It seems that you should be able to do something like this here:
```suggestion
def is_valid_shared_data_type(files):
    return all(Path(file.content_type).suffix in DATA_ENTRY_ACCEPTED_EXTENSIONS for file in files)
```

jodiezhuu: @josebui I think `.suffix` only gets the file extension so any string after `.`, I will use `.name` instead

terraso_backend/apps/shared_data/views.py
```diff
@@ -66,6 +73,12 @@ def post(self, request, **kwargs):
         return JsonResponse(data_entry.to_dict(), status=201)
 
 
+def is_valid_shared_data_type(files):
+    return all(
+        "." + Path(file.content_type).name in DATA_ENTRY_ACCEPTED_EXTENSIONS for file in files
```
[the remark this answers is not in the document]
jodiezhuu: @josebui because the content_type is of the form `[string]/[string]` (ex. `application/pdf` or `media/mp4`)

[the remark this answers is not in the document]
jodiezhuu: I can use `.suffix` if I use the file name itself and not the `content_type`, maybe this approach is better.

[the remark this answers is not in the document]
paulschreiber: The string from type/subtype does not map to file extensions. You definitely want to use `.suffix`.

terraso_backend/apps/story_map/views.py
```diff
@@ -194,13 +194,13 @@ def handle_config_media(new_config, current_config, request):
     return new_config
 
 
-def invalid_media_type(config):
+def is_valid_media_type(config):
     if "chapters" in config:
         for chapter in config["chapters"]:
             media = chapter.get("media")
             if media and not (media["type"].startswith(("image", "audio", "video", "embedded"))):
```
paulschreiber, line 201: question: do you want to use `all()` to  match the behaviour of `is_valid_shared_data_type()`?

paulschreiber (approved):

terraso_backend/apps/shared_data/views.py
```diff
@@ -52,6 +53,12 @@ def post(self, request, **kwargs):
                 context=ErrorContext(model="DataEntry", field="data_file"),
             )
             return JsonResponse({"errors": [{"message": [asdict(error_message)]}]}, status=400)
+        if not is_valid_shared_data_type(request.FILES.getlist("data_file")):
+            error_message = ErrorMessage(
+                code="Invalid media type",
```
josebui, line 58: Issue: Change the code so that it can be potentially localized in the clients side
```suggestion
                code="invalid_media_type",
```

terraso_backend/apps/story_map/views.py
```diff
@@ -117,7 +117,7 @@ def post(self, request, **kwargs):
 
         new_config = json.loads(form_data["configuration"])
 
-        if invalid_media_type(new_config):
+        if not is_valid_media_type(new_config):
```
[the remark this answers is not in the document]
paulschreiber: Do we have unit tests for this case?

Issue #623, opened 2023-06-29
Create soil data input model and associated GraphQL mutations


shrouxm:
## Description
Create a new model `SoilData` (or some better name) with a 1-1 relationship with `Site`, and columns for each data input tracking this issue. See discussion on https://github.com/techmatters/terraso-backend/issues/622 for guidance on the schema for data inputs with depth intervals.

```[tasklist]
### Tasks
- [x] Document design decisions
- [x] Make a new django app soil_id
- [x] Create a new model soil_data_inputs
- [x] Create a FK relationship to site
- [x] Add columns for the top and bottom of the depth interval
- [x] Add columns for different data inputs
- [x] Add new slope submethods
- [ ] Add GraphQL mutations for soil data
```

ltseng:
Slope user story - https://github.com/techmatters/terraso-product/issues/158

ltseng:
@shrouxm @david-code Since Jodie is no longer with us, will one of you be finishing this?

shrouxm:
@ltseng yes, i was pairing with jodie on this so i will take it over!

ltseng:
New potential data input requirement for soil texture, to be discussed next Tuesday:
- I can type the percentage of clay from 0 to 1-00% as an integer value.

PR #676, opened 2023-07-20
feat: add soil data inputs

jodiezhuu:
progression of https://github.com/techmatters/terraso-backend/issues/623


terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -23,3 +23,16 @@ class DepthDependentSoilData(BaseModel):
     soil_data_input = models.ForeignKey(SoilData, on_delete=models.CASCADE)
     depth_top = models.IntegerField(blank=True)
     depth_bottom = models.IntegerField(blank=True)
+    slope = models.IntegerField(blank=True)
```
shrouxm, line 26: for slope: 
- this one is not depth dependent, so it should be a field on `soil_data`
- for this and other fields with valid value ranges, we can use django's `MaxValueValidator` and `MinValueValidator` to ensure the percentage value is between 0 and 100: https://docs.djangoproject.com/en/4.2/ref/validators/#maxlengthvalidator
- we also need enum fields for slope shape
- for all the fields, we need to specify `null=True` (blank means that when users submit data they can omit the field, null means that the field is allowed to be null in the database. so if we had `blank=True` and `null=False`, we would need to supply a default value. but in this case we just want to allow nulls). this is true for all of the fields we're adding in this PR

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -23,3 +23,16 @@ class DepthDependentSoilData(BaseModel):
...
+    texture = models.CharField(blank=True)
```
shrouxm, line 27: for texture:
- for all the enum char fields, we should use the `choices` argument to make sure the database only accepts valid values. see e.g. the privacy field in the project model: https://github.com/techmatters/terraso-backend/blob/7d5820d221c448a72916e05de20d77402a32d2ac/terraso_backend/apps/project_management/models/projects.py#L43-L54
- there is a second enum for rock fragment volume under the soil texture spec

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -23,3 +23,16 @@ class DepthDependentSoilData(BaseModel):
...
+    color_hue_float = models.FloatField(blank=True)
+    color_hue_char = models.CharField(blank=True)
+    color_value = models.FloatField(blank=True)
+    color_chroma = models.IntegerField(blank=True)
```
shrouxm, line 31: for color:
- the wikipedia for munsell colors refers to the numeric part of hues as `substeps`, so maybe we can go with `color_hue` and `color_hue_substep`? https://en.wikipedia.org/wiki/Munsell_color_system#Hue
- same again with choices, and i think we should make them all char fields (whose values might be a string representation of a number) because these aren't actually numbers we would ever want to add together

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -23,3 +23,16 @@ class DepthDependentSoilData(BaseModel):
...
+    conductivity = models.FloatField(blank=True)
```
shrouxm, line 32: for conductivity:
- there is a requirement to restrict this field (and some other fields) to a certain number of decimal places, so it might make sense to use the `DecimalField` (and associated `DecimalValidator`) to implement that restriction
- there are enums for conductivity testing method and units

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -23,3 +23,16 @@ class DepthDependentSoilData(BaseModel):
...
+    structure = models.CharField(blank=True)
+    ph = models.FloatField(blank=True)
+    soc = models.FloatField(blank=True)
+    som = models.FloatField(blank=True)
+    sar = models.FloatField(blank=True)
+    bedrock = models.PositiveIntegerField(blank=True)
```
shrouxm, line 38: changes needed for the remaining fields are i think similar as in previous comments so i'll leave them to you!

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -23,3 +23,16 @@ class DepthDependentSoilData(BaseModel):
...
+    color_hue_float = models.FloatField(blank=True)
+    color_hue_char = models.CharField(blank=True)
+    color_value = models.FloatField(blank=True)
+    color_chroma = models.IntegerField(blank=True)
+    conductivity = models.FloatField(blank=True)
+    structure = models.CharField(blank=True)
+    ph = models.FloatField(blank=True)
+    soc = models.FloatField(blank=True)
+    som = models.FloatField(blank=True)
+    sar = models.FloatField(blank=True)
```
shrouxm, line 37: i think we should write out the full names for these fields. the acronyms are familiar to soil scientists but it might get easy to accidentally confuse them as a non-soil scientist, especially since they have the same type (accidentally passing `soc` instead of `som` or some such)

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
shrouxm, line 38: vscode has a setting called `Insert Final Newline` that will make this lint error stop happening fyi!

shrouxm (changes requested):
image

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -13,13 +13,273 @@
...
     depth_bottom = models.IntegerField(blank=True)
+    texture = models.CharField(null=True, choices=TEXTURES)
+    rock_fragment_volume = models.CharField(null=True, choices=ROCK_FRAGMENT_VOLUMES)
+    color_hue = models.CharField(null=True, choices=COLOR_HUES)
+    color_hue_substep = models.CharField(null=True, choices=COLOR_HUES_SUBSTEPS)
+    color_value = models.CharField(null=True, choices=COLOR_VALUES)
+    color_chroma = models.CharField(null=True, choices=COLOR_CHROMAS)
+    conductivity = models.DecimalField(
+        null=True, max_digits=100, decimal_places=2, validators=[MinValueValidator(0)]
+    )  ## max digit dont know
```
jodiezhuu, line 256: @shrouxm Is there a default max_digit I should use? It wasn't specified in the specs

shrouxm: you can leave `max_digits` out if you don't need it! if you have a `MaxValueValidator` on it, or if there's just no specification that the value should have an upper bound then you can just not supply the `max_digits` arg to `DecimalField`

terraso_backend/apps/soil_id/models/soil_data.py
```diff
@@ -13,11 +13,24 @@
...
+
+SLOPE_SHAPES = ((CONCAVE, _("Concave")), (CONVEX, _("Convex")), (LINEAR, _("Linear")))
+
 
 class SoilData(BaseModel):
     site = models.OneToOneField(Site, on_delete=models.CASCADE)
+    slope = models.IntegerField(
+        null=True, validators=[MinValueValidator(0), MaxValueValidator(100)]
+    )
+    slope_shape = models.CharField(null=True, choices=SLOPE_SHAPES)
```
shrouxm, line 35: nit: we need to measure slope shape across two axes, down slope and cross slope, so there should be two fields with the same choices for each axis

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -13,13 +13,273 @@
...
+
+SOIL_TESTING_METHODS = (
+    (SOIL_TESTING_DRY_COMBUSTION, _("Dry combustion")),
+    (SOIL_TESTING_WET_OXIDATION, _("Wet oxidation (Walkey-Black)")),
+    (SOIL_TESTING_LOSS_ON_IGNITION, _("Loss-on-ignition")),
+    (SOIL_TESTING_REFLECTANCE_SPECTROSCOPY, _("Reflectance spectroscopy")),
+    (SOIL_TESTING_FIELD_REFLECTOMETER, _("Field reflectometer")),
+    (OTHER, _("Other")),
+)
+
```
shrouxm, line 242: nit: we usually put the choices in the actual body of the model they're for, and place them right before the relevant field, to keep things tidier and so all the constants don't all have to be imported individually

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -13,13 +13,273 @@
...
+        max_digits=100,
+        decimal_places=1,
+        validators=[MinValueValidator(0), MaxValueValidator(100)],
+    )
+    soil_organic_carbon_testing = models.CharField(null=True, choices=SOIL_TESTING_METHODS)
+    soil_organic_matter_testing = models.CharField(null=True, choices=SOIL_TESTING_METHODS)
+    sodium_absorption_ratio = models.DecimalField(
+        null=True, max_digits=100, decimal_places=2, validators=[MinValueValidator(0)]
+    )
+    bedrock = models.PositiveIntegerField(null=True)
```
shrouxm, line 285: nit: i think 0 is a valid value for bedrock, so `IntegerField` with `MinValueValidator(0)` might make more sense

terraso_backend/apps/soil_id/models/soil_data.py
```diff
@@ -13,11 +13,24 @@
...
+    # add slope aspect ?? (north, south, east ..)
```
shrouxm, line 36: @ltseng what is the status of the slope aspect and landscape position data inputs? the product spec for them still has all that info as a suggestion, are those still in flux and we should hold off looking at them for now or can we implement according to those suggestions?

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -13,13 +13,273 @@
...
 # along with this program. If not, see https://www.gnu.org/licenses/.
 
+from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models
+from django.utils.translation import gettext_lazy as _
 
 from apps.core.models.commons import BaseModel
 from apps.soil_id.models.soil_data import SoilData
 
+OTHER = "other"
```
shrouxm, line 23: nit: i'm not sure if there's any negative implications of sharing a value between different choice fields like this? @paulschreiber do you know?

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -13,13 +13,273 @@
...
+
+RED = "R"
+YELLOW_RED = "YR"
+YELLOW = "Y"
+GREEN_YELLOW = "GY"
+GREEN = "G"
+BLUE = "B"
+BLUE_GREEN = "BG"
+
+COLOR_HUES_SUBSTEPS = (
```
shrouxm, line 83: nit: i think "2.5" etc are the substeps and "r" "yr" etc are the hues. even though we write the substep before the hue when writing it out (e.g. 2.5YR). :shrug:

shrouxm (approved):
i think this is pretty much ready, just a few more fixes. i'm approving now so i don't forget to, otherwise it won't be merge-able while i'm gone since i requested changes earlier, but i'd like to have a bit more eyes on it since there's a lot of details. i'll ask paul to review it

also @ltseng i think this code might be readable enough that you could check over if we left out any options from all the different selections? or if anything else looks wrong? but no worries if not! (none of the text here will be visible to the user, it's just for us to understand what's going on in the backend better)

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -13,13 +13,273 @@
...
+    (COLOR_CHROMA_2, _("2")),
+    (COLOR_CHROMA_3, _("3")),
+    (COLOR_CHROMA_4, _("4")),
+    (COLOR_CHROMA_5, _("5")),
+    (COLOR_CHROMA_6, _("6")),
+    (COLOR_CHROMA_7, _("7")),
+    (COLOR_CHROMA_8, _("8")),
+)
+
+CONDUCTIVITY_SATURATED_PASTE = "saturate paste"
```
ltseng, line 137: "Saturated paste," not saturate paste

terraso_backend/apps/soil_id/models/soil_data.py
ltseng: Still in flux, but I don't think I'm going to be able to dissuade BLM on those two. You can proceed with this current PR for now, and I'll file a separate issue for additions when that's ready.

ltseng (approved):
I need to finish writing the carbonates user story so we can add this in, but otherwise this looks good!

shrouxm:
oh i just remembered we haven't added any GraphQL mutations yet, but let's do that in another PR!

there also aren't any tests, but i'm not actually sure what a reasonable test suite for something like this would look like since i feel like we'd just be duplicating all this spec information? do you have thoughts @paulschreiber? never worked on something like this before

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
jodiezhuu: @shrouxm I think max_digits are required for DecimalField; I get this error if I don't have one: `soil_id.DepthDependentSoilData.soil_organic_matter: (fields.E132) DecimalFields must define a 'max_digits' attribute.`, even when I define it as none

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
jodiezhuu: Sounds good! Easier to read too!

terraso_backend/apps/soil_id/models/soil_data.py
jodiezhuu: @ltseng Just confirming, for this PR I won't include slope aspect and landscape position data inputs, is that right?

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
jodiezhuu: @shrouxm I think `PositiveIntegerField` includes 0: https://www.geeksforgeeks.org/positiveintegerfield-django-models/

shrouxm:
ok wait one more thing lol don't forget to generate the migration before merging!

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -13,13 +13,298 @@
...
+    soil_organic_carbon_testing = models.CharField(null=True, choices=SOIL_TESTING_METHODS)
+    soil_organic_matter_testing = models.CharField(null=True, choices=SOIL_TESTING_METHODS)
+
+    sodium_absorption_ratio = models.DecimalField(
+        null=True, max_digits=100, decimal_places=2, validators=[MinValueValidator(0)]
+    )
+
+    bedrock = models.PositiveIntegerField(null=True)
+
+    NONEFFERVESCENT = "noneffervescent - No bubbles form"
```
jodiezhuu, line 296: @ltseng Could you review that the carbonates data inputs are accurate? Thanks!

terraso_backend/apps/soil_id/migrations/0002_depthdependentsoildata_bedrock_and_more.py
```diff
@@ -0,0 +1,313 @@
+# Generated by Django 4.2.3 on 2023-07-27 18:48
```
paulschreiber, line 1: issue: add copyright notice.

terraso_backend/apps/soil_id/migrations/0003_depthdependentsoildata_carbonate_and_more.py
```diff
@@ -0,0 +1,108 @@
+# Generated by Django 4.2.3 on 2023-07-27 21:49
```
paulschreiber, line 1: issue: add copyright notice.

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
paulschreiber: Using OTHER twice seems fine in this case.

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -13,13 +13,298 @@
...
+    )
+    color_hue_substeps = models.CharField(null=True, choices=COLOR_HUES_SUBSTEPS)
+
+    RED = "R"
+    YELLOW_RED = "YR"
+    YELLOW = "Y"
+    GREEN_YELLOW = "GY"
+    GREEN = "G"
+    BLUE = "B"
+    BLUE_GREEN = "BG"
```
paulschreiber, line 94: question: what is the reason for abbreviating these values? Is that part of a standard notation?

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -13,13 +13,298 @@
...
 
 class DepthDependentSoilData(BaseModel):
     soil_data_input = models.ForeignKey(SoilData, on_delete=models.CASCADE)
     depth_top = models.IntegerField(blank=True)
     depth_bottom = models.IntegerField(blank=True)
+
+    SAND = "sand"
+    LOAMY_SAND = "loamy sand"
+    SANDY_LOAM = "sandy loam"
+    SILT_LOAM = "slit loam"
```
paulschreiber, line 34: issue: silt is misspelled.

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -13,13 +13,298 @@
...
+        (COLOR_CHROMA_1, _("1")),
+        (COLOR_CHROMA_2, _("2")),
+        (COLOR_CHROMA_3, _("3")),
+        (COLOR_CHROMA_4, _("4")),
+        (COLOR_CHROMA_5, _("5")),
+        (COLOR_CHROMA_6, _("6")),
+        (COLOR_CHROMA_7, _("7")),
+        (COLOR_CHROMA_8, _("8")),
+    )
+    color_chroma = models.CharField(null=True, choices=COLOR_CHROMAS)
```
paulschreiber, line 151: question: why is this an enum instead of an int with min 1 and max 8?

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -13,13 +13,298 @@
...
+    SLIGHTLY_EFFERVESCENT = "slightly effervescent - Numerous bubbles form"
+    STRONGLY_EFFERVESCENT = "strongly effervescent - Bubbles form a low foam"
+    VIOLENTLY_EFFERVESCENT = "violently effervescent - Bubbles rapidly form a thick foam"
+
+    CARBONATES = (
+        (NONEFFERVESCENT, _("Noneffervescent - No bubbles form")),
+        (VERY_SLIGHTLY_EFFERVESCENT, _("Very slightly effervescent - Few bubbles form")),
+        (SLIGHTLY_EFFERVESCENT, _("Slightly effervescent - Numerous bubbles form")),
+        (STRONGLY_EFFERVESCENT, _("Strongly effervescent - Bubbles form a low foam")),
+        (VIOLENTLY_EFFERVESCENT, _("Violently effervescent - Bubbles rapidly form a thick foam")),
```
paulschreiber, line 307: issue: use em dash, not hyphen.

paulschreiber (changes requested):

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
jodiezhuu: @ltseng What are your thoughts on this? I just took what was in the spec doc.

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
ltseng: Looks good to me!

ltseng (approved):

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
jodiezhuu: @paulschreiber I think with a small fixed set of integers like this, enums are a better choice since it ensures you can't accidentally assign incorrect or unsupported values to the field. With validators, you'd need additional checks to ensure only the allowed range of values is accepted and it only raises a [ValidationError] if the input doesn’t meet the criteria

terraso_backend/apps/soil_id/models/soil_data.py
ltseng: Right now, no. I'll write a different issue for slope aspect and landscape position later!

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
ltseng: Hue is actually a combination of two values - hue sector and hue steps, though Jon Maynard said just now that most soil scientists just call it "hue"
https://munsell.com/about-munsell-color/how-color-notation-works/munsell-hue-circle-comparison/

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
ltseng: Yes, this is part of the standard notation.

ltseng: These are part of the standard notation. https://munsell.com/about-munsell-color/how-color-notation-works/munsell-hue-circle-comparison/

paulschreiber (approved):

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
paulschreiber: Let's store this as an integer field so we can do aggregate operations (sum, average, etc.).

terraso_backend/apps/soil_id/models/depth_dependent_soil_data.py
```diff
@@ -13,13 +13,298 @@
...
+        (COLOR_VALUE_4, _("4")),
+        (COLOR_VALUE_5, _("5")),
+        (COLOR_VALUE_6, _("6")),
+        (COLOR_VALUE_7, _("7")),
+        (COLOR_VALUE_8, _("8")),
+        (COLOR_VALUE_8_5, _("8.5")),
+        (COLOR_VALUE_9, _("9")),
+        (COLOR_VALUE_9_5, _("9.5")),
+    )
+    color_value = models.CharField(null=True, choices=COLOR_VALUES)
```
paulschreiber, line 130: Let's store this as a decimal field so we can do aggregate operations (sum, average, etc.).

paulschreiber (changes requested):

paulschreiber (approved):

PR #718, opened 2023-08-01
feat: Membership List

josebui:
## Description

- Added separate app for membership code
- Connected Story Map with membership list

This first PR will just connect the new models to Story maps, the expectation is to have after this other PRs to connect this to Groups, Landscapes and Projects and also to remove the current membership logic linked to groups

### Checklist
- [x] New tests added

### Related Issues
https://github.com/techmatters/terraso-web-client/issues/1037


terraso_backend/apps/story_map/migrations/0005_storymap_membership_list.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 4.2.3 on 2023-07-31 16:16
```
paulschreiber, line 1: issue: add copyright notice.

terraso_backend/apps/collaboration/migrations/0001_initial.py
```diff
@@ -0,0 +1,113 @@
+# Generated by Django 4.2.3 on 2023-07-31 16:06
```
paulschreiber, line 1: issue: add copyright notice.

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -103,3 +108,117 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+
+    @classmethod
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        user = info.context.user
+
+        if kwargs["user_role"] != ROLE_CONTRIBUTOR:
+            logger.info(
+                "Attempt to save a Membership, but user has no permission",
+                extra=kwargs,
```
paulschreiber, line 129: question: does kwargs contain information about the user and membership list? Should the words "story map" appear in the error message?

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -103,3 +108,117 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+        if not user.has_perm(
+            StoryMap.get_perm("save_membership"),
+            obj={
+                "story_map": story_map,
+                "membership": kwargs,
+            },
+        ):
+            logger.info(
+                "Attempt to update a Membership, but user has no permission",
+                extra=kwargs,
```
paulschreiber, line 156: question: does kwargs contain information about the user and membership list? Should the words "story map" appear in the error message?

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -103,3 +108,117 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+            raise GraphQLNotAllowedException(
+                model_name=Membership.__name__, operation=MutationTypes.UPDATE
+            )
+
+        try:
+            membership = story_map.membership_list.save_member(kwargs)
+        except Membership.DoesNotExist:
+            logger.error(
+                "Attempt to update a Membership, but it was not found",
+                extra=kwargs,
```
paulschreiber, line 167: question: does kwargs contain information about the user and membership list? Should the words "story map" appear in the error message?

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -103,3 +108,117 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+                extra={"story_map_id": story_map_id, "story_map_slug": story_map_slug},
+            )
+            raise GraphQLNotFoundException(model_name=StoryMap.__name__)
+
+        try:
+            membership = story_map.membership_list.get(pk=membership_id)
+        except Membership.DoesNotExist:
+            logger.error(
+                "Attempt to delete a Membership, but it was not found",
+                extra={"membership_id": membership_id},
```
paulschreiber, line 205: question: should the words "story map" appear in the error message?

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -103,3 +108,117 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+        if not user.has_perm(
+            StoryMap.get_perm("delete_membership"),
+            obj={
+                "story_map": story_map,
+                "membership": membership,
+            },
+        ):
+            logger.info(
+                "Attempt to delete a Membership, but user has no permission",
+                extra={"user_id": user.pk, "membership_id": membership_id},
```
paulschreiber, line 218: question: should the words "story map" appear in the error message?

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -103,3 +108,117 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+
+        if not user.has_perm(
+            StoryMap.get_perm("delete_membership"),
+            obj={
+                "story_map": story_map,
+                "membership": membership,
+            },
+        ):
+            logger.info(
+                "Attempt to delete a Membership, but user has no permission",
```
paulschreiber, line 217:
```suggestion
                "Attempt to delete a membership, but user has no permission",
```

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -103,3 +108,117 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+            )
+            raise GraphQLNotAllowedException(
+                model_name=Membership.__name__, operation=MutationTypes.UPDATE
+            )
+
+        try:
+            membership = story_map.membership_list.save_member(kwargs)
+        except Membership.DoesNotExist:
+            logger.error(
+                "Attempt to update a Membership, but it was not found",
```
paulschreiber, line 166:
```suggestion
                "Attempt to update a membership, but it was not found",
```

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -103,3 +108,117 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+            )
+
+        story_map_id = kwargs["story_map_id"]
+        story_map_slug = kwargs["story_map_slug"]
+
+        try:
+            story_map = StoryMap.objects.get(slug=story_map_slug, story_map_id=story_map_id)
+        except StoryMap.DoesNotExist:
+            logger.error(
+                "Attempt to save a Membership, but Story Map was not found",
```
paulschreiber, line 142:
```suggestion
                "Attempt to save a membership, but story map was not found",
```

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -103,3 +108,117 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+        story_map_slug = graphene.String(required=True)
+
+    @classmethod
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        user = info.context.user
+
+        if kwargs["user_role"] != ROLE_CONTRIBUTOR:
+            logger.info(
+                "Attempt to save a Membership, but user has no permission",
```
paulschreiber, line 128:
```suggestion
                "Attempt to save a membership, but user has no permission",
```

terraso_backend/apps/collaboration/models/memberships.py
```diff
@@ -0,0 +1,146 @@
...
+        return User.objects.filter(id__in=member_memberships)
+
+    def has_role(self, user: User, role: str) -> bool:
+        return self.memberships.by_role(role).filter(id=user.id).exists()
+
+    def is_member(self, user: User) -> bool:
+        return self.members.filter(id=user.id).exists()
+
+    def get_membership(self, user: User):
+        return self.memberships.filter(user=user).first()
```
paulschreiber, line 90: question.should this have `approved_only()`?

terraso_backend/apps/collaboration/models/memberships.py
```diff
@@ -0,0 +1,146 @@
...
+            return cls.MEMBERSHIP_TYPE_CLOSED
+        return cls.MEMBERSHIP_TYPE_OPEN
+
+
+class MembershipObjectsManager(SafeDeleteManager):
+    def by_role(self, role):
+        return self.filter(user_role=role, membership_status=Membership.APPROVED)
+
+    def approved_only(self):
+        return self.filter(membership_status=Membership.APPROVED)
```
paulschreiber, line 108: question: can we reuse `approved_only` from groups.py?

josebui: @paulschreiber My plan is to handle this on separate stages, the plan is to eventually use the new membershipList model to get the functionality of members to Groups, Landscapes and Projects too. that is why right now there is a lot of duplication between Group and MemberhipList, but the second stage after the membership list for story maps is approved will be to connect the other 3 entities to member list and remove the membership logic coupled with group right now

terraso_backend/apps/collaboration/graphql/__init__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright © 2021-2023 Technology Matters
```
paulschreiber, line 1:
```suggestion
# Copyright © 2023 Technology Matters
```

terraso_backend/apps/collaboration/graphql/memberships.py
```diff
@@ -0,0 +1,88 @@
+# Copyright © 2021-2023 Technology Matters
```
paulschreiber, line 1:
```suggestion
# Copyright © 2023 Technology Matters
```

terraso_backend/apps/story_map/permission_rules.py
```diff
@@ -23,12 +27,56 @@ def allowed_to_view_story_map(user, story_map):
...
+    return account_membership is not None and account_membership.user_role == ROLE_COLLABORATOR
 
 
 @rules.predicate
 def allowed_to_delete_story_map(user, story_map):
     return story_map.created_by == user
 
 
+@rules.predicate
+def allowed_to_save_story_map_membership(user, obj):
```
paulschreiber, line 45: question: should this be `edit` or `update`?

terraso_backend/apps/story_map/collaboration_roles.py
```diff
@@ -0,0 +1 @@
+ROLE_COLLABORATOR = "collaborator"
```
paulschreiber, line 1: issue: add copyright notice.

terraso_backend/apps/collaboration/migrations/0002_change_contributor_to_collaborator.py
```diff
@@ -0,0 +1,15 @@
+# Generated by Django 4.2.4 on 2023-08-16 15:51
```
paulschreiber, line 1: issue: add copyright notice.

terraso_backend/apps/collaboration/models/memberships.py
```diff
@@ -0,0 +1,150 @@
+from django.core.exceptions import ValidationError
```
paulschreiber, line 1: issue: add copyright notice.

paulschreiber (changes requested):

terraso_backend/apps/story_map/permission_rules.py
josebui: @paulschreiber You mean `add` or `update` I expect this to handle both that is why I called it `save` I think I could split it if you think that is better

paulschreiber: What about `change`? We also have `allowed_to_change_group`, `allowed_to_change_landscape`, `allowed_to_delete_membership`?

josebui: That makes sense to me

terraso_backend/apps/collaboration/models/memberships.py
```diff
@@ -0,0 +1,165 @@
...
+            and membership_status == Membership.APPROVED
+        )
+
+        membership.save()
+        return is_membership_approved, membership
+
+    @property
+    def approved_members(self):
+        member_memberships = models.Subquery(self.memberships.approved_only().values("user_id"))
+        return User.objects.filter(id__in=member_memberships)
```
paulschreiber, line 100: suggestion: rename `member_memberships` to something clearer.

terraso_backend/apps/story_map/collaboration_roles.py
```diff
@@ -0,0 +1,16 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see https://www.gnu.org/licenses/.
+
+ROLE_COLLABORATOR = "collaborator"
```
paulschreiber, line 16: question: is this the same as an editor role?

paulschreiber (changes requested):

terraso_backend/apps/story_map/collaboration_roles.py
josebui: @paulschreiber yes, it changed from contributor to collaborator and the last change was to editor

paulschreiber (approved):

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -75,10 +93,29 @@ def resolve_configuration(self, info):
 
         return self.configuration
 
+    def resolve_membership_list(self, info):
+        user = info.context.user
+        if user.is_anonymous or self.membership_list is None:
+            return None
+
+        if self.created_by == user:
+            return self.membership_list
```
david-code, line 102: question: Why treat this case differently than a normal member?

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -94,12 +131,170 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
         user = info.context.user
         story_map = StoryMap.objects.get(pk=kwargs["id"])
 
-        if not user.has_perm(StoryMap.get_perm("delete"), obj=story_map):
+        if not rules.test_rule("allowed_to_delete_story_map", user, story_map):
```
david-code, line 134: question: I guess the idea here is that we should `Model.get_perm` if the permissions line up with the create, edit delete rules, and otherwise just use the `test_rule`? I ask because I think the current `Project` class should be doing the second one then.

terraso_backend/apps/graphql/schema/story_maps.py
```diff
@@ -94,12 +131,170 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+        except Exception as error:
+            logger.error(
+                "Attempt to save Story Map Memberships, but story map was not found",
+                extra={
+                    "story_map_id": story_map_id,
+                    "story_map_slug": story_map_slug,
+                    "error": error,
+                },
+            )
+            raise GraphQLNotFoundException(model_name=StoryMap.__name__)
```
david-code, line 187: suggestion: I tried adding a method to the parent class `get_or_throw` that I think you could use here.

david-code (approved):
This looks good Jose, I think this will be really easy to integrate with how Projects are currently implemented. Thanks for your work on this!

terraso_backend/apps/story_map/permission_rules.py
```diff
@@ -23,12 +27,56 @@ def allowed_to_view_story_map(user, story_map):
...
+    user_membership = obj.get("user_membership")
+
+    is_collaborator = (
+        user_membership.user_role == ROLE_COLLABORATOR
+        and user_membership.membership_status == Membership.APPROVED
+    )
+    if is_collaborator:
+        return True
+
+    return False
```
shrouxm, line 66: questions: i'm not sure i understand what is happening in
- `current_membership` is the membership being updated
- `user_membership` is the membership of the user performing the action
- i don't understand why there is a separate check for whether the user is the owner of the story map? can a user who owns a story map be given a different role than collaborator? & if a user who owns a story map has a different role than collaborator, this could says they should still be able to change memberships, but that seems confusing?
- if `current_membership` is the membership being updated, what is the check for `is_new` doing? why would we be doing a permissions check in the first place to mutate a membership that doesn't exist? or if `is_new` doesn't represent whether the membership exists yet, why aren't users allowed to change it if it's new? and why is this check after the story map owner check, so that the owner has different privileges in this regard than other collaborators?

terraso_backend/apps/collaboration/models/memberships.py
```diff
@@ -0,0 +1,167 @@
...
+    )
+
+    membership_list = models.ForeignKey(
+        MembershipList, on_delete=models.CASCADE, related_name="memberships"
+    )
+    user = models.ForeignKey(
+        User, on_delete=models.CASCADE, related_name="collaboration_memberships"
+    )
+
+    user_role = models.CharField(max_length=64)
```
shrouxm, line 148: issue: there is no validation that the role field is valid for the given membership with the current implementation, so a typo in a front end query could lead to corrupt data in the DB. this is the biggest thing that bugs me about the implementation because if we do end up wanting/needing to change it in the future (which seems very possible) it will require a data migration, all the other things will be much easier to change in code

the two ways i see to implement this as a database constraint would be:
- don't have a role field on `Membership`. models can define their own membership role/permission types by creating 1-1 fields pointing at membership. other models might not even want to have the shape of their permissions fields be a char choice field (e.g. it might be several boolean permission fields)
- turn this into a `MembershipListBase` and `MembershipBase` class and having each model that needs memberships extend them

or, without a database constraint:
- the `save_membership` function in `MembershipList` could accept an additional required parameter `valid_roles`
- continue as is, where each model using `MembershipList` must remember to do their own validation which will not be guaranteed by a DB constraint. this feels brittle to me hence the comment

shrouxm (commented):
it looks good overall! there is one bigger issue i have that i'd like to hear your thoughts on

terraso_backend/apps/collaboration/graphql/memberships.py
```diff
@@ -0,0 +1,88 @@
...
+from apps.graphql.schema.commons import TerrasoConnection
+
+from ..models import Membership, MembershipList
+
+logger = structlog.get_logger(__name__)
+
+
+class CollaborationMembershipListNode(DjangoObjectType):
+    id = graphene.ID(source="pk", required=True)
+    account_membership = graphene.Field("apps.collaboration.graphql.CollaborationMembershipNode")
```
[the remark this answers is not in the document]
josebui: You use the string when you wnat to avoid ciruclar dependencies, I think at some point in the implementation I had that issue, I will check if that is still the case, maybe I can use it directly

terraso_backend/apps/collaboration/models/memberships.py
```diff
@@ -0,0 +1,165 @@
...
+        previous_membership_status = membership.membership_status if not is_new else None
+
+        if validation_func(
+            {
+                "user_role": user_role,
+                "membership_status": membership_status,
+                "current_membership": membership,
+            }
+        ):
+            raise ValidationError("User cannot request membership")
```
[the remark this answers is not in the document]
josebui: That is a good idea, I didn't like this piece a lot, but I think providing something not inline will be cleaner I will give it a pass to see if I can improve it for this PR

terraso_backend/apps/collaboration/models/memberships.py
```diff
@@ -0,0 +1,165 @@
...
+    )
+
+    membership_list = models.ForeignKey(
+        MembershipList, on_delete=models.CASCADE, related_name="memberships"
+    )
+    user = models.ForeignKey(
+        User, on_delete=models.CASCADE, related_name="collaboration_memberships"
+    )
+
+    user_role = models.CharField(max_length=64)
```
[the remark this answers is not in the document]
josebui: Garo also has some comments about this, to avoid duplication I will add my thoughts there, but I agree an improvement will be to restrict the options here somehow, see: https://github.com/techmatters/terraso-backend/pull/718#discussion_r1316486680

terraso_backend/apps/graphql/schema/story_maps.py
josebui: You mean why don't we just create a new membership entry for the owner instead of handling the created by value?

terraso_backend/apps/graphql/schema/story_maps.py
josebui: I did some research for this and found that `has_perm` behaves diferently, it seems it actualy links to permissions that can be modified in the DB and assigned to users, so this was actually not working, with `rules.test_rule` I was sure that the validation was triggered every time, so still not completely sure about how `user.has_perm` works but with `rules.test_rule` it is calling what I need

terraso_backend/apps/graphql/schema/story_maps.py
josebui: Great, I will check that one

terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -556,6 +556,61 @@ type StoryMapNode implements Node {
...
   publishedAt: DateTime
+  membershipList: CollaborationMembershipListNode
+}
+
+type CollaborationMembershipListNode implements Node {
+  membershipType: CollaborationMembershipListMembershipTypeChoices!
+  memberships(offset: Int, before: String, after: String, first: Int, last: Int, user: ID, user_In: [ID], userRole: String, user_Email_Icontains: String, user_Email_In: [String], membershipStatus: CollaborationMembershipMembershipStatusChoices, user_Email_Not: String): CollaborationMembershipNodeConnection!
+  id: ID!
+  accountMembership: CollaborationMembershipNode
+  membershipsCount: Int
```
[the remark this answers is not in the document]
josebui: This one is a little tricky, this is something that we have also in the current group implementation, let me try to explain the issue. `memberships { totalCount }` is the total count of memberships that graph QL can return, that means that if the user can get a totalCount of 3 they should also be able to fetch the 3 memberships, also it will be linked to the filters you use for memberships (`memberships(filters) { totalCount }`) but there is also a requirement to fetch the count of memberships without giving permission to the user to actually see what those memberships are, that is what `membershipsCount` is doing, it is giving us a way to just get the number without giving access to the actual memberships data

terraso_backend/apps/collaboration/graphql/memberships.py
```diff
@@ -0,0 +1,88 @@
+# Copyright © 2023 Technology Matters
```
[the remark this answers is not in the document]
josebui: Yes, that is the idea I don't like the strucutre of moving around by type, it seems better to me if each app handles its own needed files related to graph QL inside their apps, also for this one it seemed more relevant since it can give us some clear limit to where the new collaboration code is without mixing it with the current code, and I think we will leave the graphql app (maybe we can change that name) to handle the schema definition, I will create that issue

terraso_backend/apps/graphql/schema/story_maps.py
josebui: David, I think this is a similar conversation: https://github.com/techmatters/terraso-backend/pull/718#discussion_r1317455513

terraso_backend/apps/collaboration/models/memberships.py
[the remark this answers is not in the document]
david-code: OK! I am starting to work off this branch to connect the Project to MembershipList models, and I'm using the Django [proxy inheritance](https://docs.djangoproject.com/en/4.2/topics/db/models/#proxy-models) to extend the MembershipList and add some custom code for checking the type of the role added, as well as a custom validation function. The base MembershipList could define some abstract methods (like `validate_membership` here, and maybe `validate_role`) that the child class could implement. Just an idea of what I'm thinking.

terraso_backend/apps/graphql/schema/story_maps.py
david-code: Cool, good to know!

terraso_backend/apps/collaboration/models/memberships.py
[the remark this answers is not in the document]
josebui: @david-code will that work reusing the defined GQL nodes or does that mean we will have to create specific nodes for membership list for projects?

terraso_backend/apps/graphql/schema/schema.graphql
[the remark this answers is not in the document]
shrouxm: ahh ok i see, that distinction makes sense. in the code, anonymous users always see a membership count of 0, so i guess that distinction would matter for logged in users depending whether they are or are not a member of a closed group?

terraso_backend/apps/collaboration/models/memberships.py
shrouxm: ok that sounds good, i agree DB constraints would be the most work to implement and i'm ok with just having application layer validation right now

also thinking about it more, i don't see a future situation that would _require_ a data migration if we go ahead with this implementation, since if we either wanted a DB constraint or permissions-based role model for a particular user of MembershipList, they could still implement their data extension in the way i described in the first bullet point and just ignore `Membership`'s role field. there would be a redundant column which wouldn't be elegant but not that big a deal

so i don't think this implementation would ever put us in a situation where we need to do a big refactor/data migration which is good & i feel good about the direction you're headed!

terraso_backend/apps/story_map/permission_rules.py
josebui: @shrouxm I didn't want to get that much in to the details of what ownership means at this time, because that is a bigger conversation, from some conversations with Derek and Hana ownership can be also linked to a terraso group or a landscape in the future, and that opens up a hirarchy in term of what those memberships mean, for example if the ownership of story map moves from a user to a terraso group that means that we will have a membership list linked to the terraso group and a membership list linked to the story map all (potentially) affecting what a user can do with that story map. So it still feels to me that ownership will be a separate piece not linked to membership in the long term, it feels more scalable managing ownership and membership as separate things, potentially it also seems that if the owner of an entity is another entity that has a membership list part of that ownership resolution will be linked to the membership list of the owner.

So that is why I didn't want to get that much into it and at the same time keep ownership separated so that potentially if needed it could evolve that way.

terraso_backend/apps/graphql/schema/schema.graphql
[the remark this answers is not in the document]
josebui: From what I see in the code it seems that membershipCount is not 0 for landscapes, and it also just returns the count for approved memberships, so I pontential imrpovement could be to change the name, and I think there is a bug with this count for groups, but I think that value is not being used

terraso_backend/apps/story_map/permission_rules.py
shrouxm: :face_with_spiral_eyes: :face_with_spiral_eyes: i see lol that does sound like something that should be resolved separately

PR #743, opened 2023-08-21
fix: Revert "build(deps): bump graphene-django from 3.1.3 to 3.1.5 (#733)"

david-code:
This reverts commit 4451a802a795626cf91c232fe0dfa93f761c2df5.

Seems to be the cause of https://github.com/techmatters/terraso-mobile-client/issues/291.

Will need to create bug report with project if that it the case.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....


paulschreiber:
@david-code Can we make a unit test that would have caught this failure?
@josebui Can you look this over and see if we need to make a change to be compatible with 3.1.5?

paulschreiber (approved):

PR #793, opened 2023-09-06
feat: Use MembershipList with Projects

david-code:
## Description

changes the Project class to use a MembershipList instead of a Group.

- Was able to keep most of the same code for other models that interface with Projects (I think that's basically Sites for now). 
- The GraphQL schema now exposes seperate mutations for adding a user, updating a user in the project, and removing a user from a project that does not go through the old Membership* mutations
- Created new tests for adding members to projects, removing them, etc.
- Most old tests remained more or less the same
- Had to merge the updates with project logs into it

### Checklist
- [x] New tests added

### Related Issues
Fixes #....


david-code:
Still need to merge with main and get audit log changes

terraso_backend/apps/graphql/schema/groups.py
```diff
@@ -64,10 +64,6 @@ class GroupNode(DjangoObjectType):
     account_membership = graphene.Field("apps.graphql.schema.memberships.MembershipNode")
     memberships_count = graphene.Int()
 
-    @classmethod
-    def get_queryset(cls, queryset, info):
-        return queryset.filter(project__isnull=True)
-
```
shrouxm, line 70: comment: yay :)

terraso_backend/apps/project_management/graphql/projects.py
```diff
@@ -0,0 +1,351 @@
...
+class ProjectAddUserMutation(BaseWriteMutation):
+    project = graphene.Field(ProjectNode, required=True)
+    membership = graphene.Field(MembershipNode, required=True)
+
+    model_class = Membership
+
+    class Input:
+        project_id = graphene.ID(required=True)
+        user_id = graphene.ID(required=True)
+        role = graphene.String(required=True)
```
shrouxm, line 224: suggestion: could this be an enum instead of string field? same in `UpdateUserRoleMutation`

terraso_backend/apps/project_management/permission_rules.py
```diff
@@ -56,3 +58,46 @@ def allowed_to_add_to_project(user, project):
...
+
+rules.add_rule("allowed_to_delete_user_from_project", allowed_to_delete_user_from_project)
+
+
+@rules.predicate
+def allowed_to_change_user_project_role(user, context):
+    project = context["project"]
+    requester_membership = context["requester_membership"]
+    target_membership = context["target_membership"]
+    print(requester_membership.user_role)
```
shrouxm, line 94: nit: stray print statement

terraso_backend/apps/project_management/collaboration_roles.py
```diff
@@ -0,0 +1,18 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see https://www.gnu.org/licenses/.
+
+ROLE_MANAGER = "manager"
+ROLE_CONTRIBUTOR = "contributor"
+ROLE_VIEWER = "viewer"
```
shrouxm, line 18: nit: could this be an enum?

terraso_backend/apps/project_management/graphql/projects.py
```diff
@@ -0,0 +1,351 @@
...
+            cls.not_allowed(
+                MutationTypes.UPDATE, msg="User is not allowed to change other user role"
+            )
+
+        target_membership.user_role = new_role
+        target_membership.save()
+
+        membership_updated_signal.send(sender=cls, membership=target_membership, user=requester)
+
+        return ProjectUpdateUserRoleMutation(project=project, membership=target_membership)
```
shrouxm, line 351: issue: `new_role` is never validated as being one of the possible project roles

terraso_backend/apps/project_management/migrations/0016_auto_20230915_1527.py
```diff
@@ -0,0 +1,57 @@
+# Generated by Django 4.2.5 on 2023-09-15 15:27
+
+from django.db import migrations
+
+from apps.project_management import collaboration_roles
+
+
+def migrate_forward(apps, schema_editor):
```
shrouxm, line 8: comment: cool! i have been curious how data migrations look in django

terraso_backend/apps/project_management/migrations/0017_auto_20230915_1527.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 4.2.5 on 2023-09-15 15:27
```
shrouxm, line 1: question: could these migrations all be one migration by appending their operation lists? or is there a django reason they need to be separate migrations?

terraso_backend/apps/project_management/models/projects.py
```diff
@@ -67,43 +75,78 @@ def default_settings():
...
+        user: User,
+        name: str,
+        description: Optional[str],
+        privacy: Literal["private", "public"],
+    ):
+        membership_list = cls.create_membership_list()
+        Membership.objects.create(
+            membership_list=membership_list, user=user, user_role=ROLE_MANAGER, pending_email=None
         )
+        return cls.objects.create(name=name, description=description, privacy=privacy)
```
shrouxm, line 106: nit: i think this method is unused

terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -636,7 +636,7 @@ type ProjectNode implements Node {
   updatedAt: DateTime!
   name: String!
   description: String!
-  group: GroupNode!
+  membershipList: CollaborationMembershipListNode!
```
shrouxm, line 639: nit: it would be nice if there were project specific types for `MembershipList` and `Membership` with the project model's role types, but i'm not sure how that'd work with graphene

terraso_backend/apps/project_management/graphql/projects.py
```diff
@@ -0,0 +1,351 @@
+# Copyright © 2023 Technology Matters
```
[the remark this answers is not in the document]
david-code: Cool, I didn't really think of this. Do you just move the file in one commit, and then commit changes in the next one?

terraso_backend/apps/project_management/graphql/projects.py
david-code: Oh yeah, should definitely do that.

terraso_backend/apps/project_management/migrations/0017_auto_20230915_1527.py
david-code: I followed the [migration docs](https://docs.djangoproject.com/en/4.2/howto/writing-migrations/#migrations-that-add-unique-fields) here for adding the unique `OneToOneField`. I think you're right though, they probably could be all a single migration. I think the docs suggest doing it this way to have the data migration separate from the schema migration.

terraso_backend/apps/graphql/schema/schema.graphql
david-code: Good point, I'll give this a try.

terraso_backend/apps/project_management/graphql/projects.py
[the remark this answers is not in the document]
shrouxm: i think you either make changes to the existing file and then do `git mv $OLD_PATH $NEW_PATH`, or `git mv` and then make changes, and either way should store both the move and the file changes in one commit, but there are mysterious circumstances where this doesn't work that i don't understand lol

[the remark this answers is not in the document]
david-code: OK cool, that makes sense. Maybe I'll give it a try here to test it out.

terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -614,6 +600,15 @@ enum CollaborationMembershipMembershipStatusChoices {
   PENDING
 }
 
+type CollaborationMembershipNode implements Node {
+  membershipList: ProjectMembershipListNode!
```
shrouxm, line 604: issue: should be `CollaborationMembershipListNode`

terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -556,12 +556,12 @@ type StoryMapNode implements Node {
   createdBy: UserNode
   isPublished: Boolean!
   publishedAt: DateTime
-  membershipList: CollaborationMembershipListNode
+  membershipList: ProjectMembershipListNode
```
shrouxm, line 559: issue: this shouldn't have changed

terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -556,12 +556,12 @@ type StoryMapNode implements Node {
...
 }
 
-type CollaborationMembershipListNode implements Node {
+type ProjectMembershipListNode implements Node {
   membershipType: CollaborationMembershipListMembershipTypeChoices!
-  memberships(offset: Int, before: String, after: String, first: Int, last: Int, user: ID, user_In: [ID], userRole: String, user_Email_Icontains: String, user_Email_In: [String], membershipStatus: CollaborationMembershipMembershipStatusChoices, user_Email_Not: String): CollaborationMembershipNodeConnection!
```
shrouxm, line 564: question: where did this filter go?

terraso_backend/apps/project_management/graphql/projects.py
```diff
@@ -42,10 +46,42 @@
     membership_deleted_signal,
     membership_updated_signal,
 )
+from apps.project_management import collaboration_roles
 from apps.project_management.models import Project
 from apps.project_management.models.sites import Site
 
 
+class UserRole(graphene.Enum):
```
shrouxm, line 54: suggestion: also use the enum for the mutations!

shrouxm (commented):
ugh i figured graphene was going to be really finnicky about this :sob:

terraso_backend/apps/project_management/collaboration_roles.py
david-code: I think probably, but I wanted to match the story map implementation for now. We can take another look a bit later if we want.

shrouxm:
oh no i think the fact that i merged #813 means you need to regenerate/rename the migrations i'm sorry :sob: i just did another quick look over, it all LGTM!

david-code:
> oh no i think the fact that i merged #813 means you need to regenerate/rename the migrations i'm sorry 😭 i just did another quick look over, it all LGTM!

No problem, I figured that would happen. Need to try out the migration again anyways.

PR #840, opened 2023-09-26
fix: taxonomy terms ecosystems renamed, lang codes for landscape export

josebui:
## Description

- Change taxonomy term types constants to be snake cased and upper case everywhere
- Added migration to fix ecosystem type values
- Changed language codes in landsacpe export format to use 3 characters

### Checklist
- [ ] New tests added

### Related Issues
Fixes https://github.com/techmatters/terraso-web-client/issues/1183


terraso_backend/apps/core/landscapes_views.py
```diff
@@ -64,8 +64,8 @@ def to_json(self, landscape):
                     "type": term.type,
                     "value": {
                         "original": term.value_original,
-                        "en": term.value_en,
-                        "es": term.value_es,
+                        "eng": term.value_en,
+                        "spa": term.value_es,
```
paulschreiber, line 68: question: why are we switching to ISO-639-3 here?

josebui: @paulschreiber This was changed because in the taxonomy terms languages for the orignal value we are using the 3 characters form, so for consistency we decided to change this to use the same format

paulschreiber (approved):

PR #844, opened 2023-09-29
feat: Ignore duplicate adding user to project

david-code:
For adding a user to a project, the plan is to use the project info in the redux store in the app to determine if a user belongs to a project. This would create a corner case where a user might be added to the same project twice. (Actually, I think that would exist regardless). This commit just treats that case a succesful operation and does nothing, which seems like the simplest solution to me. In the future we might want to return different error statuses if the user was already added.

### Checklist
- [x] New tests added

### Related Issues

Related somewhat to https://github.com/techmatters/terraso-mobile-client/issues/37

### DIFF
--- a/terraso_backend/apps/project_management/graphql/projects.py
+++ b/terraso_backend/apps/project_management/graphql/projects.py
@@ -315,6 +315,10 @@ class ProjectAddUserMutation(BaseWriteMutation):
             ):
                 raise ValidationError("User cannot add membership to this project")
 
+        if membership := project.get_membership(user):
+            # if user is already a member, don't do anything
+            return ProjectAddUserMutation(project=project, membership=membership)
+
         # add membership
         try:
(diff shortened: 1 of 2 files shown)


terraso_backend/apps/project_management/graphql/projects.py
```diff
@@ -315,6 +315,10 @@ def validate(context):
             ):
                 raise ValidationError("User cannot add membership to this project")
 
+        if membership := project.get_membership(user):
```
shrouxm, line 318: comment: i didn't know python had assignment expressions, cool!

shrouxm (approved):
LGTM!

was the behavior before this PR failing with an unhelpful error message because of a unique constraint violation?

david-code:
> LGTM!
> 
> was the behavior before this PR failing with an unhelpful error message because of a unique constraint violation?

Yep, that's right.

PR #845, opened 2023-09-29
feat: Add filter option to UserNode to filter users based on projects to which they belong

david-code:
Adds a filter field on the UserNode that allows a client to filter users based on if they belong to a given project. This will be used by the mobile client to query if a user can be added to a given project.

### Checklist
- [x] New tests added

### Related Issues
Related to https://github.com/techmatters/terraso-mobile-client/issues/37


terraso_backend/apps/graphql/schema/users.py
```diff
@@ -35,19 +36,30 @@
 logger = structlog.get_logger(__name__)
 
 
-class UserNode(DjangoObjectType):
-    id = graphene.ID(source="pk", required=True)
+class UserFilter(FilterSet):
+    user__in_project = CharFilter(method="filter_user_in_project")
```
shrouxm, line 40: suggestion: i used this pattern in the sites filter, it's more concise and will automagically put the right type!
```suggestion
    user__in_project = TypedFilter(field_name="collaboration_memberships__membership_list__project")
```

shrouxm, line 40: nit: i think it'd be more idiomatic/consistent with the other filter names if the filter field was just named `project` rather than `user__in_project`

shrouxm (approved):
couple suggestions but looks good!

terraso_backend/apps/graphql/schema/users.py
david-code: Nice! I missed this, this is awesome.

PR #914, opened 2023-10-23
feat: Site transfer mutation

david-code:
## Description

Adds a backend mutation for transferring sites between projects. Enforces the following privileges:

- If the site is owned by a user, the user operating the transfer must own the site. Furthermore, the user must be a contributor or manager in the project where the site is being transferred.
- If the site is being transferred from a project A to a project B, the owner must be be a manager in both projects.

Also added a utility method to cut down some of the monotony of looking up JSON paths in the graphql response.


ajabeckett (approved):
LGTM!

PR #926, opened 2023-10-27
feat: add site notes and instructions

### LINKED ISSUE #863: Add site notes and project instructions to models and GraphQL schema

ajabeckett:
## Description
Adds site notes to sites. A site can have many site notes, each with a different author. Also adds optional site instructions to project: a project can only have one site_instructions, owned by the project owner. 

### Related Issues
Fixes #863


paulschreiber:
this would be a `feat`, not a chore. (a chore is typically a non-code task.)

terraso_backend/apps/project_management/models/site_notes.py
```diff
@@ -0,0 +1,32 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see https://www.gnu.org/licenses/.
+
+from django.db import models
+
+from apps.core.models import User
+
+
+class SiteNote(models.Model):
+    site = models.ForeignKey("Site", on_delete=models.CASCADE, related_name="notes")
```
paulschreiber, line 22: question: can we import `Site` and use that? Or does it cause a circular dependency?

terraso_backend/apps/project_management/models/site_notes.py
```diff
@@ -0,0 +1,32 @@
...
+    content = models.TextField()
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+    author = models.ForeignKey(
+        User,
+        null=False,
+        blank=False,
+        on_delete=models.RESTRICT,
+        verbose_name="author to whom the note belongs",
```
paulschreiber, line 31:
```suggestion
        verbose_name="author of the note",
```

terraso_backend/apps/project_management/models/projects.py
```diff
@@ -78,6 +78,7 @@ class Meta(BaseModel.Meta):
     )
 
     seen_by = models.ManyToManyField(User, related_name="+")
+    site_instructions = models.TextField(null=True, blank=True)
```
paulschreiber, line 81: question: what is the difference between site_note and site_instructions?

terraso_backend/apps/project_management/models/site_notes.py
ajabeckett: Importing Site causes a circular dependency, unfortunately.

terraso_backend/apps/project_management/models/projects.py
ajabeckett: Many people can attach Site Notes to a site; several different data collectors might use them to make observations that aren't otherwise available in the data inputs ("I had to move because there was a tractor", "The ground was frozen", "Watch out for the goats"). 

Site Instructions, on the other hand, is a single field set on the Project that will display at the top of the Notes screen for each Site in the Project.

david-code (approved):
Looks really great! Left a few suggestions that would probably be good to fix up.

terraso_backend/tests/graphql/mutations/test_site_notes.py
```diff
@@ -0,0 +1,137 @@
...
+            errors
+        }
+    }
+"""
+
+
+def test_delete_site_note(client, site_note, user):
+    client.force_login(user)
+
+    response = graphql_query(
```
josebui, line 91: Question: Can you use `client_query` instead of `graphql_query` here? I think with that you don't have to use `client.force_login(user)`

josebui (approved):
Looks good, I just added a question

PR #945, opened 2023-10-31
fix: Set SiteNoteNode to be not null

david-code:
## Description

Needs a Terraso Connection for django-graphene to pick up on the fact that there are no null objects in the connection.


ajabeckett (approved):
Thank you!

david-code:
@ajabeckett I'll let you merge, I'm not sure if you are just using the hash from this commit directly and I don't want to interrupt you.

PR #962, opened 2023-11-08
fix: Optimized data entries query

josebui:
### Checklist
- [ ] New tests added

### Related Issues
Fixes #....


terraso_backend/apps/graphql/schema/visualization_config.py
```diff
@@ -101,17 +101,27 @@ class Meta:
 
     @classmethod
     def get_queryset(cls, queryset, info):
+        if info.field_name != "visualizationConfigs":
+            return queryset
+
```
paulschreiber, line 106: suggestion: add a comment explaining why this is only used for visualization configs.

terraso_backend/apps/graphql/schema/visualization_config.py
```diff
@@ -101,17 +101,27 @@ class Meta:
...
-            associated_groups__group__memberships__membership_status=Membership.APPROVED,
-            associated_groups__is_default_landscape_group=True,
-        ).values_list("id", flat=True)
-        all_ids = list(user_groups_ids) + list(user_landscape_ids)
-        return queryset.filter(data_entry__shared_resources__target_object_id__in=all_ids)
+
+        user_groups_ids = Subquery(
+            Group.objects.filter(
+                memberships__user__id=user_pk, memberships__membership_status=Membership.APPROVED
+            ).values("id")
```
paulschreiber, line 112: question: can you use the `approved_only()` helper here? Or does that not make sense in this context?

josebui: @paulschreiber I think we should be able to use that, let me check

josebui: @paulschreiber I checked and the `approved_only` function is linked to Membership, and in here we are using Landscape and Group. So we could potentially start from Membership but I think right now this is more effort because of the relationship of landscape to membership through the default group, maybe after the membership list changes are added this can be changed

paulschreiber (approved):

PR #1063, opened 2023-12-19
feat: project change intervals on preset change

david-code:
## Description

- Change project depth intervals when preset changes
- Delete old project intervals
- Delete old depth dependent data

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....


terraso_backend/apps/soil_id/models/project_soil_settings.py
```diff
@@ -77,6 +86,34 @@ class MeasurementUnit(models.TextChoices):
...
+    @property
+    def is_custom_preset(self):
+        return self.depth_interval_preset == DepthIntervalPreset.CUSTOM
+
+    def save(self, *args, **kwargs):
+        dirty_fields = self.get_dirty_fields()
+        with transaction.atomic():
+            result = super().save(*args, **kwargs)
+            if "depth_interval_preset" in dirty_fields or not self.id:
+                # delete project intervals...
```
paulschreiber, line 98:
```suggestion
                # delete project intervals
```

shrouxm (commented):
comment: my thinking when designing the backend data format for this was that the preset data would be stored on the client rather than updated in the backend. so the actual project depth interval rows would only represent the depth intervals for a custom depth interval selection, and when a preset is selected the client just ignores those values. this was intended to be a low implementation/UX complexity safeguard against accidental data deletion (since otherwise the only thing preventing someone from deleting all their project data by changing the preset is a confirmation dialog that can easily be mis-clicked). this definitely makes the API more consistent which is nice, but i'm not sure we should go ahead with it until we can replace that safeguard

david-code:
Superseded by #1081, closing

PR #1081, opened 2024-01-05
feat: Delete project data on a preset change

david-code:
## Description

This is just an addition on top of #1063, removing the parts that create project data intervals when the project's preset changes. This is because this data is supposed to be handled on the frontend, and only stored on the backend when the user selects the custom preset.

### Related Issues

Should fix https://github.com/techmatters/terraso-mobile-client/issues/381


terraso_backend/apps/soil_id/models/soil_data.py
```diff
@@ -205,8 +204,14 @@ class Grazing(models.TextChoices):
         WILDLIFE_GRASSLANDS = "WILDLIFE_GRASSLANDS", "Wildlife (grasslands, giraffes, ibex)"
 
     grazing_select = models.CharField(blank=True, null=True, choices=Grazing.choices)
+
+    class SoilDataDepthIntervalPreset(models.TextChoices):
+        LANDPKS = "LANDPKS"
+        NRCS = "NRCS"
+        CUSTOM = "CUSTOM"
```
paulschreiber, line 211: question: should these constants be  stored elsewhere?

terraso_backend/apps/soil_id/permission_rules.py
```diff
@@ -0,0 +1,23 @@
+# Copyright © 2021-2023 Technology Matters
```
paulschreiber, line 1:
```suggestion
# Copyright © 2024 Technology Matters
```

terraso_backend/tests/graphql/mutations/test_soil_data.py
```diff
@@ -491,6 +496,10 @@ def test_update_depth_dependent_soil_data_not_allowed(client, site):
 def test_update_project_depth_intervals(client, project_manager, project):
     client.force_login(project_manager)
 
+    # make sure there is no overlap by settings depth interval preset
+    project.soil_settings = ProjectSoilSettings(depth_interval_preset="CUSTOM")
```
paulschreiber, line 500: question: should this use a constant instead of a string?

paulschreiber (approved):

terraso_backend/apps/soil_id/models/soil_data.py
david-code: I think it's fine to put them here, following the pattern for other enums

PR #1100, opened 2024-01-11
feat: Manage soil intervals on backend

david-code:
## Description

This PR ended up doing a few things:

- Removing any old code related to creating site intervals that is no longer used
- Deleting site intervals when a site/project changes preset
- Set up the "apply_to_all" so that the interval definitions are passed from the frontend

It's basically a grab bag of fixes that I needed in order to get the frontend to work.

### Checklist
- [x] New tests added

### Related Issues
Related to https://github.com/techmatters/terraso-mobile-client/issues/502, but will still need frontend work


terraso_backend/apps/graphql/schema/sites.py
```diff
@@ -301,6 +318,9 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
                 old_projects.append(site.project)
 
         Site.bulk_change_project(to_change, project)
+        # update site depth intervals where necessary
```
paulschreiber, line 321: issue: Please explain more about what "when necessary" means. Under what circumstances does this occur?

terraso_backend/apps/soil_id/models/project_soil_settings.py
```diff
@@ -90,20 +92,63 @@ class MeasurementUnit(models.TextChoices):
...
 
+    @property
+    def methods(self):
+        field_names = [
+            field.name.removesuffix("_enabled")
+            for field in SoilDataDepthInterval._meta.fields
+            if field.name.endswith("_enabled")
+        ]
+        return {
+            field_name + "_enabled": getattr(self, field_name + "_required")
```
paulschreiber, line 103:
```suggestion
           f"{field_name}_enabled": getattr(self, f"{field_name}_required")
```

terraso_backend/apps/soil_id/models/project_soil_settings.py
```diff
@@ -90,20 +92,63 @@ class MeasurementUnit(models.TextChoices):
...
 
+    def convert_site_intervals_to_preset(
+        self, new_preset: Optional[DepthIntervalPreset] = None, sites: Optional[list[Site]] = None
+    ):
+        if not sites:
+            sites = Site.objects.filter(project=self.project, soil_data__isnull=False)
+        intervals = SoilDataDepthInterval.objects.filter(soil_data__site__in=sites)
+
+        intervals.delete()
+        # create new site intervals
```
paulschreiber, line 132:
```suggestion

        # create new site intervals
```

terraso_backend/apps/soil_id/models/project_soil_settings.py
```diff
@@ -90,20 +92,63 @@ class MeasurementUnit(models.TextChoices):
...
+            DepthIntervalPreset.LANDPKS: LandPKSIntervalDefaults,
+            DepthIntervalPreset.NRCS: NRCSIntervalDefaults,
+        }
+        preset = new_preset or self.depth_interval_preset
+        interval_bounds = options.get(preset)
+        if interval_bounds:
+            intervals = []
+            for site in sites:
+                if not hasattr(site, "soil_data"):
+                    # Not going to create an interval if soil_data doesn't exist
```
paulschreiber, line 143:
```suggestion
                    # Do not create an interval if soil_data doesn't exist
```

terraso_backend/apps/soil_id/models/soil_data.py
```diff
@@ -214,6 +214,13 @@ class SoilDataDepthIntervalPreset(models.TextChoices):
         choices=SoilDataDepthIntervalPreset.choices, blank=True, null=True
     )
 
+    def remove_from_project(self):
+        SoilDataDepthInterval.objects.filter(soil_data=self).delete()
+        # Default is to set the site interval to custom for now
+        # At some point, user will be able to set a default preset
```
paulschreiber, line 220:
```suggestion
        # TODO: At some point, user will be able to set a default preset
```

paulschreiber (commented):
1. Please have @shrouxm review as well.
2. We cannot rely on soft_delete-d items being available, as these records get removed periodically.

terraso_backend/apps/soil_id/models/project_soil_settings.py
```diff
@@ -90,20 +92,63 @@ class MeasurementUnit(models.TextChoices):
...
                 DepthDependentSoilData.delete_in_project(self.project.id)
+                # create site intervals
+                self.convert_site_intervals_to_preset(new_preset)
         return result
 
+    def convert_site_intervals_to_preset(
+        self, new_preset: Optional[DepthIntervalPreset] = None, sites: Optional[list[Site]] = None
+    ):
+        if not sites:
+            sites = Site.objects.filter(project=self.project, soil_data__isnull=False)
```
shrouxm, line 128: question: we leave out sites with null soil data because we'll call this method again when the soil data is created for those sites?

trying to brainstorm cases where that might cause an issue: when we query a site that doesn't yet have soil data, a default soil data object is created for the site, which would not have the project's depth intervals. does that cause problems with how you're thinking about things?

shrouxm: oh i see that you added a method to create the soil data object when a site is created. then we have this unvalidated assumption that the soil data objects will exist, which will be true for all new sites for now, but might cause issues down the line

terraso_backend/apps/soil_id/graphql/soil_data.py
```diff
@@ -235,11 +235,20 @@ class Input:
         sodium_adsorption_ratio_enabled = graphene.Boolean()
         soil_structure_enabled = graphene.Boolean()
 
+        new_depth_interval = graphene.Field(DepthIntervalInput)
```
shrouxm, line 238: question: what is the intention/use of the `new_depth_interval` arg? it seems to be for mutating an existing depth interval's start and end values, but i don't understand why we would want to do that

shrouxm (commented):
i haven't had time to think through all the possibilities yet to see if there's any bugs in the implementation, but i am very nervous about the amount of backend data chasing there is with this approach. i'm trying to understand the problem a bit more, in the original thread you asked the question: 

"This is what is currently causing me a headache - say there is a project level interval. What do we do when the user activates the allow input switch? The project intervals aren't saved on the backend / don't have the input information if they are." 

reading through the code for the update depth interval mutation again, it seems like the answer to that question would be "a depth interval object will be created on demand to store the configuration". is that wrong, & if not is there some other issue that that creates down the line?

also, re this question from this PR:

> it's not obvious to me how to differentiate between "inactive" intervals linked to the old preset, and "active" intervals linked to the new preset.

> I think one implementation could be change the object manager that the soil depth interval uses. It would automatically filter out any intervals that don't match the current preset in any database query.

we could implement that logic clientside as well

maybe we should do a quick call to think through this i'm feeling very :face_with_spiral_eyes: about it

terraso_backend/apps/graphql/schema/sites.py
david-code: removed

paulschreiber (approved):

PR #1148, opened 2024-02-08
fix: don't delete users when cleaning up old models

paulschreiber:
## Description
Don't delete users when cleaning up old models. Preserves integrity of audit log.


josebui (approved):
Suggestion: Add new test for this

PR #1188, opened 2024-03-06
fix: depth interval default methods

shrouxm:
## Description
Some small changes to the data model to handle subtle cases of depth interval default methods. Previously we were only storing true/false for whether an input method was enabled, now we also allow the input method to be unset. This is because there are actually three possible states for input methods being enabled: explicit yes, explicit no, and unspecified. I think there may be some even longer tail UX edge cases to consider, but this data model lets us handle more situations with less complexity created by manual data chasing.

### Checklist

### Related Issues
Related to https://github.com/techmatters/terraso-mobile-client/issues/502


paulschreiber (approved):

terraso_backend/apps/graphql/schema/sites.py
```diff
@@ -25,6 +25,7 @@
 from apps.project_management.graphql.projects import ProjectNode
 from apps.project_management.models import Project, Site, sites
 from apps.soil_id.models import SoilData
+from apps.soil_id.models.project_soil_settings import ProjectSoilSettings
```
paulschreiber, line 28: issue: this import causes a linter error. should we remove it, or do we need to make a change to use it?

PR #1195, opened 2024-03-11
feat: Set is first login value inside token

josebui:
## Description

- Added parameter inside token to know if the user was created with a service

### Checklist
- [x] New tests added

### Related Issues
- https://github.com/techmatters/terraso-web-client/issues/1145


paulschreiber (changes requested):
Please rename `created_with_service` to something that indicates it is true only for new years and false for existing users. Please also add a comment to that effect.

paulschreiber (approved):

PR #1206, opened 2024-03-20
feat: consolidate enum formats

shrouxm:
## Description
Removes a redundant field and uses enums more consistently for project/soil id code.

### Related Issues
Needed for https://github.com/techmatters/terraso-client-shared/pull/359

### Verification steps
Still need to fix some tests.


terraso_backend/apps/project_management/models/projects.py
```diff
@@ -37,24 +33,19 @@ class Meta(BaseModel.Meta):
 
 
 class ProjectMembership(Membership):
-    """A proxy class created soley for graphene schema reasons"""
+    """A proxy class created solely for graphene schema reasons"""
```
paulschreiber, line 36: question: what are those schema reasons?

shrouxm: i don't know, i was just fixing a typo in the doc comment. my guess is it creates an opaque wrapper type around `Membership` allowing us to evolve the concept of a `ProjectMembership` separately from a `LandscapeMembership`, `StoryMapMembership`, etc without needing to make breaking changes for clients

terraso_backend/apps/project_management/graphql/projects.py
```diff
@@ -55,30 +55,17 @@
...
 from apps.soil_id.models.project_soil_settings import ProjectSoilSettings
 
-
-class UserRole(graphene.Enum):
-    viewer = collaboration_roles.ROLE_VIEWER
-    contributor = collaboration_roles.ROLE_CONTRIBUTOR
-    manager = collaboration_roles.ROLE_MANAGER
+ProjectRoleEnum = graphene.Enum(
+    "ProjectManagementProjectRoleChoices", [(c[0], c[1]) for c in ProjectRole.choices]
+)
```
tm-ruxandra, line 60: Can we have a comment explaining what this code is doing? Even within the confines of graphene APIs it seems pretty opaque.

shrouxm: they have an API to do this! updated the code. and my opaque tomfoolery was causing a bug anyway so good catch

paulschreiber (approved):

PR #1222, opened 2024-04-03
fix: New user's language will be set to English; profile language dropdown gets populated

knipec:
## Description
Previously the language was getting set to "true", so the language dropdown in the Account Profile page was not getting populated with a default value. 
The backend code was sending that value, so change it to the default language in settings.py (which currently seems to just be English without more complicated logic)

Also rename the user preference constants to not be prefixed with "notification".

### Checklist
- [x] Verified on desktop

### Related Issues
Fixes #https://github.com/techmatters/terraso-web-client/issues/1630 

### Verification steps
See https://github.com/techmatters/terraso-web-client/issues/1630


terraso_backend/apps/auth/services.py
```diff
@@ -82,8 +82,16 @@ def sign_up_with_microsoft(self, authorization_code):
...
     def _set_default_preferences(self, user):
-        for notification_key in NOTIFICATION_KEYS:
-            UserPreference.objects.create(user=user, key=notification_key, value="true")
+        for notification_key in USER_PREFS_KEYS:
+            default_val = "true"
+            if notification_key == USER_PREFS_KEY_LANGUAGE:
+                # Reformat from 'en-us' to 'en-US'. Assumes language code comes in as 'xx-XX'
+                def format(lang_code):
+                    parts = lang_code.split("-")
+                    return f"{parts[0]}-{parts[1].upper()}"
```
paulschreiber, line 91: suggestion: move this function to a utilities file so we can reuse it.

paulschreiber (approved):

terraso_backend/apps/auth/services.py
```diff
@@ -29,11 +29,12 @@
 from django.utils import timezone
 
 from apps.core.models import UserPreference
-from apps.core.models.users import NOTIFICATION_KEYS
+from apps.core.models.users import USER_PREFS_KEY_LANGUAGE, USER_PREFS_KEYS
 from apps.storage.services import ProfileImageService
 
 from .providers import AppleProvider, GoogleProvider, MicrosoftProvider
 from .signals import user_signup_signal
+from .utils import format_to_upper
```
paulschreiber, line 37: suggestion: what about using core_formatters instead of creating a new file? I would also give it a more specific name, like `uppercase_locale`.

knipec: Ooh, didn't see that file existed, thanks!

PR #1223, opened 2024-04-03
fix: add Django UserAdmin

### LINKED ISSUE #191: Add ability to change user password from Django admin
## Description
Add ability to change user password from Django admin.

paulschreiber:
## Description
Add Django UserAdmin. Allows changing passwords in admin ui.

Fixes #191.


knipec (approved):
Nice, thanks!

PR #1254, opened 2024-04-25
fix: prevent the removal of the only manager from projects

tm-ruxandra:
## Description
Add check to reject project user-deletion requests if the user in question is the only manager for the project. The system was accepting incoming user-removal requests for the project's only manager, despite requirements that removing the only manager on a project is not permitted.

### Checklist
- [x] New tests added

### Related Issues
Fixes #[496](https://github.com/techmatters/terraso-mobile-client/issues/496)

### Verification steps

See automated tests covering these cases.

### DIFF
--- a/terraso_backend/apps/project_management/graphql/projects.py
+++ b/terraso_backend/apps/project_management/graphql/projects.py
@@ -398,6 +398,10 @@ class ProjectDeleteUserMutation(BaseWriteMutation):
             cls.not_allowed(
                 MutationTypes.DELETE, msg="User not allowed to remove member from project"
             )
+
+        if project.is_sole_manager(user):
+            cls.not_allowed(MutationTypes.DELETE, msg="Cannot remove the last manager from project")
+
         # remove membership
         membership = project.get_membership(user)
         membership.delete()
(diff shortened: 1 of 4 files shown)


terraso_backend/apps/project_management/graphql/projects.py
```diff
@@ -398,6 +398,10 @@ def mutate_and_get_payload(cls, root, info, project_id, user_id):
             cls.not_allowed(
                 MutationTypes.DELETE, msg="User not allowed to remove member from project"
             )
+
+        if project.is_sole_manager(user):
+            cls.not_allowed(MutationTypes.DELETE, msg="Cannot remove the last manager from project")
```
paulschreiber, line 403: question: should we use rules.test_rule(), like we do for landscape and groups? cc @josebui

terraso_backend/tests/project_management/models/test_projects.py
```diff
@@ -0,0 +1,43 @@
+# Copyright © 2023 Technology Matters
```
paulschreiber, line 1:
```suggestion
# Copyright © 2024 Technology Matters
```

knipec (approved):

terraso_backend/apps/project_management/graphql/projects.py
tm-ruxandra: The enforcement mechanisms for validation are generally inconsistent across the mutation code and I did not want to introduce indirection + dependence on a library for an extremely straightforward check here. For example, just above this we have similar checks like `if not target_membership:` and `if not requester_membership:`.

Issue #1266, opened 2024-05-01
feat: define data schema for site soil matches GraphQL query 


shrouxm:
## Description
Depends on #1265. Create a second endpoint which additionally accepts soil data inputs, and returns a bit more metadata for the ranking. This corresponds to the site soil ID screens in the Figma. See #1265 for further details.

Issue #1265, opened 2024-05-01
feat: define data schema for non-site soil match GraphQL query


shrouxm:
Define a new field on our GraphQL query object with the expected input/output of the soil ID algorithm.

As far as I'm aware, the capri figma screens are the best source of truth for what data we need, and the `list.json` and `rank.json` files on the `feat/soilid` branch of the backend are the best source of truth for what data the algorithm outputs.

PR #1273, opened 2024-05-02
feat: soil id endpoints

shrouxm:
## Description
Defines GraphQL endpoints for the soil id calls, with dummy data for now. Still need to add tests to validate that the endpoints work/return data with the right shape.

### Checklist
- [ ] New tests added

### Related Issues
Progress on #1265 and #1266

### DIFF
--- a/terraso_backend/apps/graphql/schema/schema.graphql
+++ b/terraso_backend/apps/graphql/schema/schema.graphql
@@ -1547,9 +1547,11 @@ type LocationBasedSoilMatch {
 
 """Provides information about soil at a particular location."""
 type SoilInfo {
+  dataSource: String!
   soilSeries: SoilSeries!
   ecologicalSite: EcologicalSite
   landCapabilityClass: LandCapabilityClass!
+  soilData: [SoilIdSoilData!]!
 }
 
 """Information about a soil series."""
@@ -1558,7 +1560,6 @@ type SoilSeries {
   taxonomySubgroup: String!
   description: String!
   fullDescriptionUrl: String!
-  dataSource: String!
 }
 
 """Information about an ecological site."""
@@ -1572,6 +1573,15 @@ type EcologicalSite {
 type LandCapabilityClass {
   capabilityClass: String!
   subClass: String!
+}
+
+type SoilIdSoilData {
+  depthInterval: DepthInterval!
+  texture: SoilIdDepthDependentSoilDataTextureChoices
+  rockFragmentVolume: SoilIdDepthDependentSoilDataRockFragmentVolumeChoices
+  colorHue: Float
+  colorValue: Float
+  colorChroma: Float
 }
 
 """
--- a/terraso_backend/apps/graphql/schema/schema.graphql
+++ b/terraso_backend/apps/graphql/schema/schema.graphql
@@ -90,6 +90,7 @@ type Query {
     orderBy: String
   ): AuditLogNodeConnection
   sharedResource(shareUuid: String!): SharedResourceNode
+  soilId: SoilID!
 }
 
 type GroupNode implements Node {
@@ -1527,6 +1528,96 @@ String, Boolean, Int, Float, List or Object.
 """
 scalar GenericScalar
 
+"""Soil ID algorithm queries."""
+type SoilID {
+  locationBasedSoilMatches(latitude: Float!, longitude: Float!): LocationBasedSoilMatches!
+  dataBasedSoilMatches(latitude: Float!, longitude: Float!, data: SoilIDInputData!): DataBasedSoilMatches!
+}
+
+"""A ranked group of soil matches based solely on a coordinate pair."""
+type LocationBasedSoilMatches {
+  matches: [LocationBasedSoilMatch]!
+}
+
+"""A soil match based solely on a coordinate pair."""
+type LocationBasedSoilMatch {
+  soilInfo: SoilInfo!
+  match: SoilMatch!
+}
+
+"""Provides information about soil at a particular location."""
+type SoilInfo {
+  soilSeries: SoilSeries!
+  ecologicalSite: EcologicalSite
+  landCapabilityClass: LandCapabilityClass!
+}
+
+"""Information about a soil series."""
+type SoilSeries {
+  name: String!
+  taxonomySubgroup: String!
+  description: String!
+  fullDescriptionUrl: String!
+  dataSource: String!
+}
+
+"""Information about an ecological site."""
+type EcologicalSite {
+  name: String!
+  id: String!
+  url: String!
+}
+
+"""Caveat: may want to update these fields to an enum at some point."""
+type LandCapabilityClass {
+  capabilityClass: String!
+  subClass: String!
+}
+
+"""
+The likelihood score and rank within the match group for a particular soil type.
+"""
+type SoilMatch {
+  score: Float!
+  rank: Int!
+}
+
+"""
+A ranked group of soil matches based on a coordinate pair and soil data.
+"""
+type DataBasedSoilMatches {
+  matches: [DataBasedSoilMatch]!
+}
+
+"""A soil match based on a coordinate pair and soil data."""
+type DataBasedSoilMatch {
+  soilInfo: SoilInfo!
+  locationMatch: SoilMatch!
+  dataMatch: SoilMatch!
+  combinedMatch: SoilMatch!
+}
+
+"""Soil data provided to the soil ID algorithm."""
+input SoilIDInputData {
+  slope: SoilIdSoilDataSlopeSteepnessSelectChoices
+  intervals: [SoilIDInputDepthDependentData]!
+}
+
+"""Depth dependent data provided to the soil ID algorithm."""
+input SoilIDInputDepthDependentData {
+  depthInterval: DepthIntervalInput!
+  texture: SoilIdDepthDependentSoilDataTextureChoices = null
+  rockFragmentVolume: SoilIdDepthDependentSoilDataRockFragmentVolumeChoices = null
+  colorHue: Float
+  colorValue: Float
+  colorChroma: Float
+}
+
+input DepthIntervalInput {
+  start: Int!
+  end: Int!
+}
+
 type Mutations {
   addGroup(input: GroupAddMutationInput!): GroupAddMutationPayload!
   addLandscape(input: LandscapeAddMutationInput!): LandscapeAddMutationPayload!
@@ -2191,11 +2282,6 @@ input DepthDependentSoilDataUpdateMutationInput {
   sodiumAbsorptionRatio: Decimal
   carbonates: SoilIdDepthDependentSoilDataCarbonatesChoices
   clientMutationId: String
-}
-
-input DepthIntervalInput {
-  start: Int!
-  end: Int!
 }
 
 type SoilDataUpdateDepthIntervalMutationPayload {
--- a/terraso_backend/apps/graphql/schema/schema.graphql
+++ b/terraso_backend/apps/graphql/schema/schema.graphql
@@ -256,7 +256,7 @@ input SoilIdInputDepthDependentData {
   depthInterval: DepthIntervalInput!
   texture: SoilIdDepthDependentSoilDataTextureChoices = null
   rockFragmentVolume: SoilIdDepthDependentSoilDataRockFragmentVolumeChoices = null
-  colorLab: LABColorInput = null
+  colorLAB: LABColorInput = null
 }
 
 input DepthIntervalInput {
--- a/terraso_backend/apps/graphql/schema/schema.graphql
+++ b/terraso_backend/apps/graphql/schema/schema.graphql
@@ -155,9 +155,7 @@ type SoilIdDepthDependentData {
   depthInterval: DepthInterval!
   texture: SoilIdDepthDependentSoilDataTextureChoices
   rockFragmentVolume: SoilIdDepthDependentSoilDataRockFragmentVolumeChoices
-  colorHue: Float
-  colorValue: Float
-  colorChroma: Float
+  munsellColorString: String
 }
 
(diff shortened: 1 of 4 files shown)


terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -1527,6 +1528,96 @@ String, Boolean, Int, Float, List or Object.
 """
 scalar GenericScalar
 
+"""Soil ID algorithm queries."""
+type SoilID {
+  locationBasedSoilMatches(latitude: Float!, longitude: Float!): LocationBasedSoilMatches!
+  dataBasedSoilMatches(latitude: Float!, longitude: Float!, data: SoilIDInputData!): DataBasedSoilMatches!
```
tm-ruxandra, line 1534: I'm a little concerned about the complexity of the SoilIDInputData type -- would it make sense to take in a site ID instead, and then look up the necessary data in the query resolver?

I definitely think it's good to have the underlying implementation be decoupled from any notion of sites, but I'm pretty sure the only use-case we have for data-based lookups will be sourcing this from sites, right? And so it might be better for the UI to not have to transform and send this data over the wire when it's already available to the backend.

(If I'm wrong about any of this, please correct me!)

shrouxm: your understanding is all correct, and it's a good question. this is my thinking for why to construct the API this way:
- in either case, we need to implement the transformation from site to soil id input data in one place (either in the frontend or the backend), so it's not necessarily more work either way
- the amount of data sent over the wire is very small in practice (it's a list of length ~10 of json objects with ~5 fields)
- i agree that for the capri feature set, it seems a bit nicer to just take a site ID. but there are multiple important future design goals in the app which would require us to eventually do this work client-side anyway:
  - running the algorithm offline
  - running the algorithm for non-signed in users who don't want to persist their data on our backend

i try to stay away from pre-designing things, but in this case where the trade-off seems small and the future considerations are pretty concrete and important it seemed like the better choice. thoughts?

tm-ruxandra: Your reasoning here makes sense and I think that ultimately I'd like to have both versions of the endpoint for exactly the reasons given here. I think I'm coming at this from a slight preference for backend-side transformation of data for the reason that our backend seems more well-tested / easier to nail down data conditions, but it's ultimately a decision of low consequence (we can always add a backend API built on top of this existing one if it proves too cumbersome to do the transformation on the frontend.)

terraso_backend/apps/soil_id/graphql/soil_id.py
knipec, line 1: Nitpick: Might be nice to have them mimic the same order as the schema.graphql file :)

terraso_backend/apps/soil_id/graphql/soil_id.py
```diff
@@ -0,0 +1,211 @@
...
+
+class SoilIDInputDepthDependentData(graphene.InputObjectType):
+    """Depth dependent data provided to the soil ID algorithm."""
+
+    depth_interval = graphene.Field(DepthIntervalInput, required=True)
+    texture = graphene.Field(DepthDependentSoilDataNode.texture_enum())
+    rock_fragment_volume = graphene.Field(DepthDependentSoilDataNode.rock_fragment_volume_enum())
+    color_hue = graphene.Float()
+    color_value = graphene.Float()
+    color_chroma = graphene.Float()
```
knipec, line 101: Question: I'm slightly surprised these are represented as Floats, is that just how the soil color algorithm works?

terraso_backend/apps/graphql/schema/schema.graphql
knipec, line 1: Question: In the Figma, there are separate conditions for "No soil map data (NOTCOM)" and "Soil map data not published (NOTPUB)" ([Figma](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=6685-22317&mode=design&t=xRjNvHJFcdBTCsbS-0, [GitHub #1228](https://github.com/techmatters/terraso-mobile-client/issues/1228))

What, if anything yet, does that correspond to in the API?

terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -1527,6 +1528,96 @@ String, Boolean, Int, Float, List or Object.
...
+type EcologicalSite {
+  name: String!
+  id: String!
+  url: String!
+}
+
+"""Caveat: may want to update these fields to an enum at some point."""
+type LandCapabilityClass {
+  capabilityClass: String!
+  subClass: String!
```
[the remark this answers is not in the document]
paulschreiber: Data Source: can be either `SSURGO` or `STATSGO`.

terraso_backend/apps/soil_id/graphql/soil_id.py
paulschreiber: My understanding is the YR is basically an enum. It could be an integer.

shrouxm: my understanding/thinking on this:

- hue gets split into 10 steps, being `R YR Y GY G BG B PB P RP`, and each step is split into 4 substeps, being `2.5 5 7.5 10`
- value and chroma when displayed are rounded to nearest soil color (or to nearest integer if they're far away from nearest soil color)
- if chroma is close to 0, instead the whole thing is displayed as `N $VALUE/`

we store them as floats because the only time we actually need them in munsell notation format is for user display, but all internal manipulation of the data is easier in numerical space, and we also avoid erasing information unnecessarily this way (e.g. it might be helpful for iterating on the color algorithm to see what the unrounded outputs are)

terraso_backend/apps/soil_id/graphql/soil_id.py
shrouxm: good point!

terraso_backend/apps/graphql/schema/schema.graphql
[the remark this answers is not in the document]
shrouxm: data source is on the `SoilSeries` type up above

terraso_backend/apps/graphql/schema/schema.graphql
shrouxm: this API doesn't reflect those cases since i wasn't sure what we were going for for MVP yet. if i were to implement it, i would probably add a `NoSoilDataReason` enum with these cases, and then the `locationBasedSoilMatches` would return `LocationBasedSoilMatches | NoSoilDataReason` or something similar to that. `dataBasedSoilMatches` would be a bit more finnicky since we still return data in that case, but some subset of the data would be missing

terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -1527,6 +1528,96 @@ String, Boolean, Int, Float, List or Object.
...
+  soilSeries: SoilSeries!
+  ecologicalSite: EcologicalSite
+  landCapabilityClass: LandCapabilityClass!
+}
+
+"""Information about a soil series."""
+type SoilSeries {
+  name: String!
+  taxonomySubgroup: String!
+  description: String!
```
tm-ruxandra, line 1559: Question - will these be 18n key strings, or actual copy text embedded in the dataset?

shrouxm: mmmm very good point, i will file a follow-up issue about this. jon's algorithm code outputs these descriptions as english text, but i think (though am not sure) that his code fetches them from a DB column somewhere else. on the other hand, having read several of the descriptions now they seem to be code-generated very simplistically from pieces of data about the soil. so i don't think there's anything to do now, but we will definitely need to figure out a better way to navigate this when localizing

terraso_backend/apps/soil_id/graphql/soil_id.py
knipec: Ok, I still don't entirely understand how to translate the floats into munsell notation, but as long as you know how to, I'm good with this :)

terraso_backend/apps/graphql/schema/schema.graphql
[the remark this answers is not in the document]
shrouxm: (i now moved it onto `SoilInfo`)

terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -1527,6 +1528,106 @@ String, Boolean, Int, Float, List or Object.
...
+  colorHue: Float
+  colorValue: Float
+  colorChroma: Float
+}
+
+"""
+The likelihood score and rank within the match group for a particular soil type.
+"""
+type SoilMatch {
+  score: Float!
```
tm-ruxandra, line 1591: I think we need another piece of data here indicating whether the match is coming from inside the soil map unit or some adjacent unit. (This is specified in the design for the match info sheet).

shrouxm: ooh good call!

terraso_backend/apps/soil_id/graphql/soil_id.py
shrouxm: there is [a function for that](https://github.com/techmatters/terraso-mobile-client/blob/ada733f533d008d1b02738aaa26cba949556be63/dev-client/src/screens/SoilScreen/ColorScreen/utils/munsellConversions.ts#L201-L215) and [a longer explanation](https://github.com/techmatters/terraso-mobile-client/blob/ada733f533d008d1b02738aaa26cba949556be63/dev-client/src/screens/SoilScreen/ColorScreen/utils/munsellConversions.ts#L18-L42) in the file `munsellConversions.ts`

terraso_backend/apps/soil_id/graphql/soil_id.py
shrouxm: turns out this is not possible because the logic the schema codegen uses is to define parent objects first, then child objects. but in the python code, if you define the parent object first, there will be an error because the child object doesn't exist yet :( so they're sort of forced to have opposite orders

terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -92,6 +94,180 @@ type Query {
...
+
+"""A ranked group of soil matches based solely on a coordinate pair."""
+type LocationBasedSoilMatches {
+  matches: [LocationBasedSoilMatch!]!
+}
+
+"""A soil match based solely on a coordinate pair."""
+type LocationBasedSoilMatch {
+  dataSource: String!
+  inMapUnit: Boolean!
```
knipec, line 111: Question (non-blocking for this PR though!): I wasn't entirely clear if this was a boolean or an enum when Courtney mentioned it. Is it just "in soil map unit" or "near soil map unit", and if your location is deemed too far away then it would be one of the "no map data" cases (at least post-Capri)? 

(Please validate this stream of consciousness lol: I was at first like "what should we show in Capri here if it's a 'no map data' case" -- but actually that's not applicable because you can only to the screen that displays that info if you have a non-empty set of soil matches, which requires there being map data ... I think?)

shrouxm: i don't fully understand this either, it's on my todo list to look into this a bit more carefully. @CourtneyLee333 could you expand on the intention/source of the requirement for the "In map unit" text in the figma?

tm-ruxandra: I think this is the last issue I see with this API, so once this gets squared away I think we're good to merge.

terraso_backend/tests/graphql/test_soil_id.py
```diff
@@ -0,0 +1,190 @@
...
+    }
+  }
+"""
+    + SOIL_MATCH_FRAGMENTS
+)
+
+
+def test_location_based_soil_matches_endpoint(client):
+    response = graphql_query(
+        LOCATION_BASED_MATCHES_QUERY, variables={"latitude": 0.0, "longitude": 0.0}, client=client
```
knipec, line 91: Question: Would this fail once we have actual data coming in? I don't think there's soil at null island? :)

knipec (approved):
I'm down for you to merge this and sort out the inMapUnit stuff in a follow-up PR

terraso_backend/tests/graphql/test_soil_id.py
shrouxm: i think you're right :thinking: https://www.openstreetmap.org/#map=10/0.0000/0.0000

& yes, this will get updated to real coordinates when there's real data!

shrouxm:
@knipec i just pushed one more change, incorporating the map unit distance changed we talked about in stand-up, and i also changed the format of the color inputs we were talking about earlier in the review based on my attempts at wiring this up to the actual algorithm:
- it turns out the lowest friction thing will be to return the munsell color string as is from the algorithm, 
- and the algorithm expects color to be input to it in Lab format, which we already functions for converting to munsell on the frontend

terraso_backend/apps/soil_id/graphql/soil_id.py
```diff
@@ -118,15 +116,19 @@ class DataBasedSoilMatches(graphene.ObjectType):
     matches = graphene.List(graphene.NonNull(DataBasedSoilMatch), required=True)
 
 
+class LabColorInput(graphene.InputObjectType):
```
knipec, line 119: Suggestion: Does "Lab" stand for something? If so, maybe make a comment or name them more clearly?

shrouxm: nope, it's just LAB :shrug:  https://en.wikipedia.org/wiki/CIELAB_color_space

paulschreiber: suggestion: call this `LABColorInput`, since we capitalized `RGB` elsewhere.

PR #1276, opened 2024-05-03
feat: soil id algorithm dependency (but with Ubuntu)

### LINKED ISSUE #1267: discovery: attempt to integrate soil ID algorithm code into backend codebase
## Description
The soil ID algorithm code is currently defined in a separate repo, and we need to run it out of our django backend. We could either try to turn the `soil-id-algorithm` repo into a python package, or inline the code into the backend. I'm writing this as a discovery task to have a time-based check-in about how it's going in whenever we attempt to do this.

paulschreiber:
## Description
- Use osgeo docker image for newer GDAL.
- Add soil id algorithm module dependency
- Download  soil id data via GitHub actions
- Run basic `list_soils` test

## Related issue
This fixes #1267.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,8 +33,7 @@ jobs:
           make build DC_ENV=ci
 
       - name: Run lint checks
-        run: |
-          make lint
+        run: make lint
 
   build-and-test:
     runs-on: ubuntu-latest
@@ -72,5 +71,4 @@ jobs:
           gdown 1K0GkqxhZiVUND6yfFmaI7tYanLktekyp # DBF
 
       - name: Run tests using built Docker image
-        run: |
-          make test-ci DC_ENV=ci
+        run: make test-ci DC_ENV=ci
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,6 +50,26 @@ jobs:
           echo "SENTRY_DSN=${{ secrets.SENTRY_DSN }}" >> .env
           make build DC_ENV=ci
 
+      - name: Restore data cache
+        uses: actions/cache@v4
+        with:
+          path: Data/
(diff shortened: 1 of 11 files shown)


shrouxm (changes requested):
when i check out this branch and try to run the test i get
```
FAILED terraso_backend/tests/soil_id/test_soil_id.py::test_soil_list - FileNotFoundError: [Errno 2] No such file or directory: 'Data/LandPKS_munsell_rgb_lab.csv'
```

could you add documentation for how to get this working with local dev? & i think ideally the data would get loaded in the docker image itself so that we're getting consistent data everywhere we're running the app

shrouxm:
update:
- i got it to work locally by copying the gdown commands from the CI file
- it seems only the shp file is onerously large, so could we check all of the other ones into git?
- i looked a bit into how to deploy it, it seems the only straightforward way to do it without having to redownload the 500mb file every time we deploy is to use render's "persistent volumes" feature
- i think we should sync on this and talk through it

shrouxm:
writing up some more thoughts / thoughts that existed in scattered convos/syncs:
- re: the small files: i realized another thing we could do is check them into the soil-id-algorithm repo so that downstream dependencies don't even need to worry about them, which seems like the mostest ideal solution. i think having fewer files that we need to manually shuffle around between every machine that runs this code is more of a benefit than any downside of not managing the algorithm's static assets in a totally consistent way, since checking things into git is so straightforward that there's not any mental overhead for doing it, just being aware that it's set up that way
- re: persistent volumes, that seems good for now, but would definitely be nice to manage these file dependencies in a declarative way down the road
- re: the dockerfile, we could try pinning `soil-id-algorithm`'s python gdal dependency to an earlier version

paulschreiber:
Replaced with #1290.

Issue #1112, opened 2024-01-22
Add backend APIs for user deletion requests


ltseng:
Backend work to allow users to send a request to delete their accounts. See parent issue for details. 

## References
- [Product Spec](https://docs.google.com/document/d/18t3V8KWm1wft7K66f91wwKfUQdwIQqJonFJszGaHGeM/edit#heading=h.b4eice3hjq0r)

## Acceptance Criteria 

- [ ] Write a script that allows staff members to complete deletion. Script should
  - [ ] delete the user account
     - [ ] delete unaffiliated sites
     - [ ] delete projects and affiliated sites where the user is the sole manager
  - [ ] Unsubscribe users from LandPKS mailing list https://app.hubspot.com/contacts/9151742/lists/16/filters
  - [ ] Use the hubspot template to inform the user their account has been deleted - [Delete Account Email Template](https://app.hubspot.com/templates/9151742)

CourtneyLee333:
@paulschreiber I don't know how to test this one.

PR #1278, opened 2024-05-06
feat: Create account deletion form via hubspot API

paulschreiber:
## Description
Create account deletion form via hubspot API.

### Related Issues
First part of #1112.


tm-ruxandra (commented):
This implementation seems simple enough and makes sense to me as written. Do I understand correctly that some other future work will expose this functionality to the frontend for the delete button? That's my only question, to be sure that this covers the intended ticket scope

paulschreiber:
@tm-ruxandra correct. we will need to add two GraphQL endpoints
-a mutation to set a user as "deletion request initiated"
- a query to get the status of a deletion request

tm-ruxandra (approved):

PR #1286, opened 2024-05-07
feat: add elevation to sites model

paulschreiber:
## Description
Add elevation to sites model.

## Related tickets
Part of https://github.com/techmatters/terraso-product/issues/722


terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -1958,6 +1959,7 @@ input SiteAddMutationInput {
   name: String!
   latitude: Float!
   longitude: Float!
+  elevation: Float!
```
shrouxm, line 1962: comment: i don't think it's a blocker, but we should file a follow-up tech-debt issue to maintain awareness of it, it seems like there's a bit of shtuff to think through re whether the client will always be able to supply an elevation. like if our elevation API provider goes down no new sites can be created, or if a user creates a site offline we need to insert a step where we compute the elevation when we get back online. but again i think this is ok for now

terraso_backend/apps/project_management/migrations/0028_site_elevation.py
```diff
@@ -0,0 +1,34 @@
...
+
+    dependencies = [
+        ("project_management", "0027_project_measurement_units"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="site",
+            name="elevation",
+            field=models.FloatField(default=0.0),
```
shrouxm, line 31: question: is this migration in sync with the code? i didn't expect to see the `default=0.0` here?

shrouxm (approved):

Issue #1269, opened 2024-05-02
feat: hook up location based soil matches graphql query to real soil ID algorithm


shrouxm:
## Description
Depends on #1265 and #1267. Size reflects that it's not clear whether this will require making changes to the actual soil ID algorithm.

PR #1289, opened 2024-05-09
feat: wire up soil id algorithm

shrouxm:
## Description
returns actual soil id algorithm data from our graphql endpoint!

this depends on both #1273 and #1290 so it's based on a merge of the two branches, this will need to be cleaned up before merging this

### Checklist

### Related Issues
Related to #1269

### DIFF
--- a/terraso_backend/apps/soil_id/graphql/soil_id.py
+++ b/terraso_backend/apps/soil_id/graphql/soil_id.py
@@ -203,6 +203,10 @@ sample_soil_infos = [
 ]
 
 
+def resolve_texture(texture: str):
+    return texture.upper().replace(" ", "_")
+
+
 def resolve_rock_fragment_volume(rock_fragment_volume: int):
     if rock_fragment_volume <= 1:
         return DepthDependentSoilData.RockFragmentVolume.VOLUME_0_1
@@ -224,7 +228,7 @@ def resolve_soil_data(soil_match):
     for id, bottom_depth in bottom_depths.items():
         depth_dependent_data[int(id)] = SoilIdDepthDependentData(
             depth_interval=DepthInterval(start=prev_depth, end=bottom_depth),
-            texture=soil_match["texture"][id].upper().replace(" ", "_"),
+            texture=resolve_texture(soil_match["texture"][id]),
             rock_fragment_volume=resolve_rock_fragment_volume(soil_match["rock_fragments"][id]),
             munsell_color_string=soil_match["munsell"][id],
         )
@@ -269,13 +273,13 @@ def resolve_soil_match(soil_match):
 
 
 def resolve_location_based_soil_matches(_parent, _info, latitude: float, longitude: float):
-    algorithm_result = list_soils(lat=latitude, lon=longitude, plot_id=None, site_calc=False)
+    result = list_soils(lat=latitude, lon=longitude, plot_id=None, site_calc=False)
 
-    if isinstance(algorithm_result, str):
+    if isinstance(result, str):
         return None
 
     matches = []
-    for match in algorithm_result["soilList"]:
+    for match in result["soilList"]:
         if match["id"]["rank_loc"] != "Not Displayed":
             matches.append(resolve_soil_match(match))
     return LocationBasedSoilMatches(matches=matches)
--- a/terraso_backend/apps/soil_id/graphql/soil_id.py
+++ b/terraso_backend/apps/soil_id/graphql/soil_id.py
@@ -14,12 +14,14 @@
 # along with this program. If not, see https://www.gnu.org/licenses/.
 
 import graphene
+from soil_id.us_soil import list_soils
 
 from apps.soil_id.graphql.soil_data import (
     DepthDependentSoilDataNode,
     DepthInterval,
     DepthIntervalInput,
 )
+from apps.soil_id.models.depth_dependent_soil_data import DepthDependentSoilData
 
 
 class EcologicalSite(graphene.ObjectType):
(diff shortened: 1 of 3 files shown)


terraso_backend/apps/soil_id/graphql/soil_id.py
```diff
@@ -201,26 +203,84 @@ class SoilIdInputData(graphene.InputObjectType):
...
+
+def resolve_soil_data(soil_match):
+    bottom_depths = soil_match["bottom_depth"]
+    prev_depth = 0
+    depth_dependent_data = [None] * len(bottom_depths)
+
+    for id, bottom_depth in bottom_depths.items():
+        depth_dependent_data[int(id)] = SoilIdDepthDependentData(
+            depth_interval=DepthInterval(start=prev_depth, end=bottom_depth),
+            texture=soil_match["texture"][id].upper().replace(" ", "_"),
```
paulschreiber, line 227: suggestion: if we'll use this elsewhere, move to a helper function.

terraso_backend/apps/soil_id/graphql/soil_id.py
```diff
@@ -201,26 +203,84 @@ class SoilIdInputData(graphene.InputObjectType):
...
+    site_data = soil_match["site"]["siteData"]
+    ecological_site = soil_match["esd"]["ESD"]
+    if ecological_site["ecoclassid"] == "":
+        ecological_site = None
+    else:
+        ecological_site = EcologicalSite(
+            name=ecological_site["ecoclassname"],
+            id=ecological_site["ecoclassid"],
+            url=ecological_site["esd_url"],
+        )
```
paulschreiber, line 248:
```suggestion
    if ecological_site["ecoclassid"]:
        ecological_site = EcologicalSite(
            name=ecological_site["ecoclassname"],
            id=ecological_site["ecoclassid"],
            url=ecological_site["esd_url"],
        )
    else:
        ecological_site = None
```

terraso_backend/apps/soil_id/graphql/soil_id.py
```diff
@@ -201,26 +203,84 @@ class SoilIdInputData(graphene.InputObjectType):
...
+
+    return LocationBasedSoilMatch(
+        data_source=site_data["dataSource"],
+        distance_to_nearest_map_unit_m=site_data["minCompDistance"],
+        match=SoilMatchInfo(score=soil_id["score_loc"], rank=int(soil_id["rank_loc"]) - 1),
```
paulschreiber, line 253: question: are we using different indexing for rank in the algorithm code? would it make sense to fix it there?

terraso_backend/apps/soil_id/graphql/soil_id.py
```diff
@@ -201,26 +203,84 @@ class SoilIdInputData(graphene.InputObjectType):
...
             ),
-        ]
+            ecological_site=ecological_site,
+            soil_data=resolve_soil_data(soil_match),
+        ),
     )
 
 
+def resolve_location_based_soil_matches(_parent, _info, latitude: float, longitude: float):
+    algorithm_result = list_soils(lat=latitude, lon=longitude, plot_id=None, site_calc=False)
```
paulschreiber, line 272:
```suggestion
    result = list_soils(lat=latitude, lon=longitude, plot_id=None, site_calc=False)
```

terraso_backend/apps/soil_id/graphql/soil_id.py
```diff
@@ -201,26 +203,84 @@ class SoilIdInputData(graphene.InputObjectType):
...
+
+    if isinstance(algorithm_result, str):
```
paulschreiber, line 274:
```suggestion
    if isinstance(result, str):
```

terraso_backend/apps/soil_id/graphql/soil_id.py
```diff
@@ -201,26 +203,84 @@ class SoilIdInputData(graphene.InputObjectType):
...
+        return None
+
+    matches = []
+    for match in algorithm_result["soilList"]:
```
paulschreiber, line 278:
```suggestion
    for match in result["soilList"]:
```

paulschreiber (approved):

terraso_backend/apps/soil_id/graphql/soil_id.py
shrouxm: there are soooo many little things i'd rather change upstream, but for now i'm trying to treat it as a static dependency as much as possible since the code is so brittle i feel like it'd be easy to accidentally introduce subtle bugs

terraso_backend/apps/soil_id/graphql/soil_id.py
shrouxm: i prefer being explicit about what i'm checking for in conditions

terraso_backend/apps/soil_id/graphql/soil_id.py
paulschreiber: It'd be pretty easy to add this code to the soil-id-algorithm repo. Shall we do that?

shrouxm:
@paulschreiber ok, so if i merge this now then main will be broken on deploys if anyone calls these endpoints (which could be happening sooner than later). are you planning to set up the render backend deployments to support running the algorithm or do you want me to do that? (otherwise i'll start looking into hooking up the data-based soil matches query)

terraso_backend/apps/soil_id/graphql/soil_id.py
shrouxm: i don't see any reason to spend the effort right now, and i can easily see it becoming a rabbit hole. even if i just ctrl+f for "rank_loc" i get 12 different hits with multiple different cases, and then it'd also be weird if we didn't update rank_data and rank_loc_data to be consistent, which is another 30 or so locations, and making sure we don't fudge any of those numbers when this code already works feels unnecessary right now when we know we're going to have to do a much bigger overhaul of the code upstream anyway

paulschreiber:
@shrouxm I added the CSV and shapefile data to `/var/data` on production, staging and testing.

shrouxm:
@paulschreiber we still need a way to tell the algorithm to look in that folder for the data though right?

shrouxm:
oh nvm i see that env var is set up. ok i will deploy this to staging and check if i can make the api query ok!

PR #1290, opened 2024-05-10
feat: soil id algorithm dependency (with Debian 12)

### LINKED ISSUE #1267: discovery: attempt to integrate soil ID algorithm code into backend codebase

paulschreiber:
## Description
- Use Debian Testing PPA for newer GDAL.
- Add soil id algorithm module dependency
- Download  soil id data via GitHub actions
- Run basic `list_soils` test

## Related issue
This fixes #1267.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,30 @@
-FROM python:3.12.3-slim-bullseye
-
-RUN adduser --disabled-password terraso
+FROM ghcr.io/osgeo/gdal:ubuntu-full-latest
 
 ENV PATH /home/terraso/.local/bin:$PATH
 # see https://github.com/aws/aws-cli/tags for list of versions
 ENV AWS_CLI_VERSION 2.8.12
 
 RUN apt-get update && \
+    apt-get install -q -y --no-install-recommends software-properties-common
+
+# prevent tzdata from prompting
+RUN ln -fs /usr/share/zoneinfo/America/New_York /etc/localtime
+
+RUN apt-get update && \
     apt-get install -q -y --no-install-recommends \
+                     python3.12 python3.12-dev \
                      build-essential libpq-dev dnsutils libmagic-dev mailcap \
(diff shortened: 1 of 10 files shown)


Dockerfile
```diff
@@ -1,11 +1,17 @@
...
 
 ENV PATH /home/terraso/.local/bin:$PATH
 # see https://github.com/aws/aws-cli/tags for list of versions
 ENV AWS_CLI_VERSION 2.8.12
 
+# Add testing sources and pin the GDAL packages to testing
+# Allows us to get 3.8.x versions of GDAL
+RUN sed 's/bookworm/testing/g' /etc/apt/sources.list.d/debian.sources >  /etc/apt/sources.list.d/testing.sources
+
+RUN echo 'Package: libgdal-dev gdal-bin\nPin: release a=testing\nPin-Priority: 900' > /etc/apt/preferences.d/gdal-testing
```
shrouxm, line 13: question: is there a more structured `apt` command to do this rather than manually inserting text into `apt` config files?

shrouxm (commented):
yay much cleaner! one more question

Dockerfile
paulschreiber: I looked, and couldn't find one.

shrouxm: ok, i also don't understand why we aren't using gdal 3.6.x to avoid these incompatibilities like we discussed last week but this seems fine for now

shrouxm (approved):

Dockerfile
paulschreiber: Everyone's dev environment has 3.8.x, so I wanted to match that.

PR #1331, opened 2024-06-21
fix: ensure no default value for USER_PREFS_KEY_ACCOUNT_DELETION

paulschreiber:
## Description
Ensure no default value is set for USER_PREFS_KEY_ACCOUNT_DELETION upon account createion.

### DIFF
--- a/terraso_backend/apps/auth/services.py
+++ b/terraso_backend/apps/auth/services.py
@@ -30,7 +30,11 @@ from django.utils import timezone
 
 from apps.core.formatters import uppercase_locale
 from apps.core.models import UserPreference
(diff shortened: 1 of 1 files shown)


terraso_backend/apps/auth/services.py
```diff
@@ -87,6 +91,8 @@ def _set_default_preferences(self, user):
             default_val = "true"
             if notification_key == USER_PREFS_KEY_LANGUAGE:
                 default_val = uppercase_locale(settings.DEFAULT_LANGUAGE_CODE)
+            elif notification_key == USER_PREFS_KEY_ACCOUNT_DELETION:
+                continue
```
tm-ruxandra, line 95: Why continue, rather than "false"?

tm-ruxandra (approved):

PR #1343, opened 2024-06-24
build: update geopandas from 0.14.4 to 1.0.0

paulschreiber:
## Description
Update geopandas from 0.14.0 to 1.0.0.


paulschreiber:
@josebui this breaks some tests (and presumably some functionality). Please take a look and update our code for compatibility with the new version.

paulschreiber:
This is the same error we get when installing fiona from a binary:
https://github.com/techmatters/terraso-backend/actions/runs/10706204143/job/29683592776

PR #1329, opened 2024-06-19
feat: update and test soil id algorithm

shrouxm:
## Description
Updates the soil ID algorithm and adds unit tests for our resolver code.

### Checklist
- [x] New tests added

### Related Issues
Progress on https://github.com/techmatters/soil-id-algorithm/issues/8

PR #1344, opened 2024-06-25
feat: enable soil ID algorithm

shrouxm:
## Description
Re-enables the soil ID algorithm. Stacked on #1329, and should be merged to main after #1329 is merged so that it's easy to revert.

### DIFF
--- a/terraso_backend/apps/soil_id/graphql/soil_id/resolvers.py
+++ b/terraso_backend/apps/soil_id/graphql/soil_id/resolvers.py
@@ -18,10 +18,6 @@ from typing import Optional
 from soil_id.us_soil import list_soils, rank_soils
 
 from apps.soil_id.graphql.soil_data import DepthInterval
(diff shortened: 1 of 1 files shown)


knipec (approved):
Haven't looked at the upstream #1329 change but assuming that's good then 👍 on this one! 

However, if I just get this branch, and try playing with the app: 
- All sites I have fail to show the soil id LocationPrediction in the LocationDashboard
- Most temporary locations also fail to retrieve soil ID info for the map callout, instead giving console error "Failed to parse Terraso API error response One or more required values are NA. TypeError: messages.map is not a function (it is undefined)"
Screenshot 2024-06-25 at 6 22 17 PM

knipec (approved):
Re-approving! 
- I had gotten the backend repo changes but hadn't updated the dependencies so didn't have changes to the soil-id-algorithm
- Some of my sites still fail, but the same lat/long will fail as a temporary location too

PR #1345, opened 2024-06-26
feat: refactor soil id endpoints to return more specific error messages

shrouxm:
## Description
Refactors the soil ID algorithm endpoints to always return successfully but provide a specific reason if no results are returned. For now we will not be more granular than saying either "data unavailable" or "algorithm failed", since that would require refactoring the actual algorithm to return more useful output.'

Related to https://github.com/techmatters/terraso-product/issues/895

### DIFF
--- a/terraso_backend/apps/soil_id/graphql/soil_id/resolvers.py
+++ b/terraso_backend/apps/soil_id/graphql/soil_id/resolvers.py
@@ -13,9 +13,11 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see https://www.gnu.org/licenses/.
 
+import traceback
 from typing import Optional
 
-from soil_id.us_soil import list_soils, rank_soils
+import structlog
+from soil_id.us_soil import SoilListOutputData, list_soils, rank_soils
 
 from apps.soil_id.graphql.soil_data import DepthInterval
 from apps.soil_id.graphql.soil_id.schema import (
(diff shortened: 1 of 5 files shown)


terraso_backend/apps/soil_id/graphql/soil_id/resolvers.py
```diff
@@ -124,21 +130,35 @@ def resolve_location_based_soil_match(soil_match: dict):
...
+
+        failure = resolve_list_output_failure(list_output)
+        if failure is not None:
+            return failure
 
-    return resolve_location_matches_from_soil_id_result(result.soil_list_json)
+        return resolve_location_based_soil_matches(list_output.soil_list_json)
+    except Exception:
+        logger.error(traceback.format_exc())
+        return SoilIdFailure(reason=SoilIdFailureReason.ALGORITHM_FAILURE)
```
knipec, line 161: Question: I imagine there's no meaningful difference between this error case and the other case that returns `ALGORITHM_FAILURE`, just that in this one we'll get a log about it?

knipec (approved):

terraso_backend/apps/soil_id/graphql/soil_id/resolvers.py
shrouxm: hmm good question, i should log a different error for the other one, it's a catch-all because i feel confused by what the soil ID algorithm can return, but i should still log the confusion so it's easier to track down if that code path is getting triggered

PR #1383, opened 2024-07-25
fix: correct depth degree and percent validation

paulschreiber:
## Description
Correct depth degree and percent validation. Matches frontend changes in https://github.com/techmatters/terraso-mobile-client/pull/1834, which addresses https://github.com/techmatters/terraso-mobile-client/issues/848.


terraso_backend/apps/soil_id/migrations/0016_alter_soildata_slope_steepness_degree_and_more.py
```diff
@@ -0,0 +1,53 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see https://www.gnu.org/licenses/.
+
+# Generated by Django 5.0.7 on 2024-07-25 20:24
+
+import django.core.validators
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
tm-ruxandra, line 22: Is this safe to run on a live database? What will the effect be if existing records violate this constraint?

paulschreiber: We don't have any custom depth intervals outside this range in the database.

tm-ruxandra (approved):

PR #1413, opened 2024-08-13
fix: update esd_url to edit_url

paulschreiber:
## Description
update esd_url to edit_url for compatibility with new soil id endpoint.


terraso_backend/apps/soil_id/graphql/soil_id/resolvers.py
knipec, line 1: What's the meaning of "edit" in "edit_url"? Like, it's not a url that allows the user to edit stuff, right?

knipec (approved):

terraso_backend/apps/soil_id/graphql/soil_id/resolvers.py
paulschreiber: EDIT = Ecosystem Dynamics Interpretive Tool.

Issue #1487, opened 2024-10-08
debt: remove individual soil data update queries


shrouxm:
## Description
Once we have implemented pushing data to the server in a bulk sync operation, the `updateSoilData`, `updateDepthDependentSoilData`, and depending what order we do things in also the `updateSoilDataDepthInterval` and `deleteSoilDataDepthInterval` will all be unused. Once clients are up to date with the new app version, we can remove them from the codebase.

shrouxm:
superseded by #1528

PR #1536, opened 2024-11-05
fix: enum fields in soil data push

shrouxm:
## Description
Fixes the bug with enum fields in the soil data push mutation, and takes some steps towards more robust testing.

Related to the testing part of #1487.

### DIFF
--- a/terraso_backend/apps/soil_id/models/soil_data_history.py
+++ b/terraso_backend/apps/soil_id/models/soil_data_history.py
@@ -13,11 +13,21 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see https://www.gnu.org/licenses/.
 
+import enum
+
+from django.core.serializers.json import DjangoJSONEncoder
 from django.db import models
 
 from apps.core.models import User
 from apps.core.models.commons import BaseModel
 from apps.project_management.models.sites import Site
+
+
+class JSONEncoder(DjangoJSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, enum.Enum):
+            return obj.value
+        return super().default(obj)
 
 
 # NOTE: this table may contain data associated with sites that was submitted
@@ -55,4 +65,4 @@ class SoilDataHistory(BaseModel):
     #     }
     #   }]
     # }
-    soil_data_changes = models.JSONField()
+    soil_data_changes = models.JSONField(encoder=JSONEncoder)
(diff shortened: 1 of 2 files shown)


terraso_backend/apps/soil_id/models/soil_data_history.py
```diff
@@ -13,13 +13,23 @@
...
 from apps.core.models import User
 from apps.core.models.commons import BaseModel
 from apps.project_management.models.sites import Site
 
 
+class JSONEncoder(DjangoJSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, enum.Enum):
+            return obj.value
+        return super().default(obj)
```
shrouxm, line 30: comment: i don't _think_ this is super specific to the `SoilDataHistory` model, but i'm not sure where to put it or if other models which use it would need to implement it differently so i'm keeping it local for now

tm-ruxandra: I think this is fine for now -- if the pattern reoccurs we can extract something more general.

terraso_backend/tests/graphql/mutations/test_soil_data.py
```diff
@@ -904,17 +953,31 @@ def test_push_soil_data_success(client, user):
     assert response.json()
     result = response.json()["data"]["pushSoilData"]
     assert result["errors"] is None
-    assert result["results"][0]["result"]["soilData"]["slopeAspect"] == 10
+    print(response.json())
```
tm-ruxandra, line 956: We should remove this stray print.

terraso_backend/tests/graphql/mutations/test_soil_data.py
```diff
@@ -875,7 +877,54 @@ def test_push_soil_data_success(client, user):
...
+        "carbonates": "SLIGHTLY_EFFERVESCENT",
+    }
+
+    nested_changes = {
+        "depthDependentData": [
+            {"depthInterval": {"start": 0, "end": 10}, **depth_dependent_changes}
+        ],
         "depthIntervals": [
             {
                 "depthInterval": {"start": 0, "end": 10},
```
tm-ruxandra, line 930: Should we also exhaustively test depth interval fields?

tm-ruxandra (changes requested):
See comments -- looks good, two notes on testing that I think we should address.

tm-ruxandra (approved):

PR #1613, opened 2025-01-16
fix: lcc resolver

shrouxm:
## Description
Fix the soil ID request resolver to handle some edge cases in the algorithm's return values for LCC.

### Checklist
- [x] New tests added

### Related Issues
Related to https://github.com/techmatters/soil-id-algorithm/issues/119


knipec (approved):

terraso_backend/apps/soil_id/graphql/soil_id/resolvers.py
knipec, line 1: I assume this means we'll just show an empty string as the value on the frontend?

shrouxm: on the front end it shows "Not available" when it's an empty string!

PR #1616, opened 2025-02-18
fix: use uv instead of standard pip

paulschreiber:
## Description
Use uv instead of standard pip. Speeds up make lock/make lock dev at least 10X.


Makefile
```diff
@@ -135,10 +135,10 @@ download-soil-data:
 	gdown 1K0GkqxhZiVUND6yfFmaI7tYanLktekyp \
 
 ${VIRTUAL_ENV}/scripts/black:
-	pip install black
+	uv pip install black
```
knipec, line 138: Question/suggestion: Glancing at [the docs](https://docs.astral.sh/uv/concepts/projects/layout/#the-pyprojecttoml), it says “It is not recommended to modify the project environment manually, e.g., with uv pip install” but in our makefile we do that in a few places – should it use `uv add` as the docs suggest? If not, could you clarify this for me?

knipec (commented):
General: 
1. Issue: Looks like this is causing the lint step to fail
2. Question: Does this have any implications for how we'd use anything in this repo? I assume mostly we'll just use the same make commands and the same input/outputs, but is there anything else to be aware of?
3. Could we chat about this on a call so I can confirm I understand?

Makefile
paulschreiber: For now, we're using the backwards compatible method of replacing pip with `uv pip`. Later, we can move the requirements to the toml file and use `uv sync`.

requirements.txt
```diff
@@ -1,11 +1,5 @@
-#
-# This file is autogenerated by pip-compile with Python 3.13
-# by the following command:
-#
-#    make lock
-#
---no-binary fiona
```
knipec, line 7: Question: Is this removed because it's already in base.in?

Makefile
```diff
@@ -133,12 +135,3 @@ download-soil-data:
 	gdown 185Qjb9pJJn4AzOissiTz283tINrDqgI0; \
 	gdown 1P3xl1YRlfcMjfO_4PM39tkrrlL3hoLzv; \
 	gdown 1K0GkqxhZiVUND6yfFmaI7tYanLktekyp \
-
-${VIRTUAL_ENV}/scripts/black:
-	pip install black
-
-${VIRTUAL_ENV}/scripts/isort:
-	pip install isort
-
```
knipec, line 142: Question: Were these deleted for this change, or for the `ruff` PR?

Dockerfile
knipec, line 1: Question: Could you explain the changes in this file?

Dockerfile.dev
knipec, line 1: Question: Could you explain why this is needed now when it wasn't before?

Makefile
knipec, line 1: Reposting question: Does this PR have any implications for how we'd use anything in this repo? I assume mostly we'll just use the same make commands and the same input/outputs, but is there anything else to be aware of?

knipec (commented):
Mostly just questions trying to understand what's going on

Makefile
paulschreiber: I decided to delete them here instead of converting them to `uv pip install` and then remove them in the `ruff` PR.

requirements.txt
paulschreiber: In `base.in`, I had to move `--no-binary fiona` to its own line. https://github.com/astral-sh/uv/issues/4420#issuecomment-2179502987

Dockerfile
paulschreiber: When running the GitHub actions, we pass `--system` to the `uv pip install` commands to avoid creating an unnecessary virtual environment. That's why `DC_ENV=ci` is passed.

The other change is to install `uv` itself.

Dockerfile.dev
paulschreiber: Not really. I'm not sure why it worked before.

Dockerfile
knipec: Hmm, I was asking about `Dockerfile` -- it sounds like that explanation is about the `build.yml` file?

knipec (approved):

Dockerfile
paulschreiber: In the `Dockerfile` — from `uv` documentation

- the `COPY` line gets `uv` installed
- the container needs a virtual environment
- no longer install `pip` since it's not used

PR #1619, opened 2025-02-20
Fix: Editing depth interval should remove old interval earlier

knipec:
## Description
**Previously**: Editing a depth interval could cause a conflict if the new depth interval overlapped the old one:
```{"data":{"pushSoilData":{"results":[{"siteId":"6652e310-0938-48c6-99f8-b2895183004e","result":{"__typename":"SoilDataPushEntryFailure","reason":"INVALID_DATA"}}],"errors":null}}}```
This is because it would try to add the new interval to the site, validate the site's intervals (where it could fail), and then delete the old interval. This would cause the old interval to stay around, though a new interval would also appear, and the project conflict dialog would show.

**Now**: Delete the old interval before trying to add the new interval to the site. Honestly I didn't investigate why the project conflict dialog would show, but it no longer does. Also add a test.

### Related Issues
Addresses #1618

### Verification steps
See bug

### DIFF
--- a/terraso_backend/apps/soil_id/graphql/soil_data/push_mutation.py
+++ b/terraso_backend/apps/soil_id/graphql/soil_data/push_mutation.py
@@ -196,6 +196,8 @@ class SoilDataPush(BaseWriteMutation):
                     reason=reason,
                 )
 
+            # Delete depth intervals first to avoid errors during validation if new
+            # intervals overlap with deleted ones
             SoilDataPush.delete_depth_intervals(soil_data, deleted_depth_intervals)
             SoilDataPush.update_soil_data(soil_data, update_data)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by paulschreiber — APPROVED** [review]

File: terraso_backend/apps/soil_id/graphql/soil_data/push_mutation.py [hunk 1]
```diff
@@ -196,10 +196,10 @@ def mutate_and_get_entry_result(
                     reason=reason,
                 )
 
+            SoilDataPush.delete_depth_intervals(soil_data, deleted_depth_intervals)
```
paulschreiber, line 199: suggestion: add a comment indicating deleting needs to happen first (and why).

PR #1622, opened 2025-02-20
fix: use ruff instead of flake8, black and isort

paulschreiber:
## Description
use ruff instead of flake8, black and isort.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,9 +32,8 @@ clean:
 createsuperuser: check_rebuild
 	$(DC_RUN_CMD) python terraso_backend/manage.py createsuperuser
 
-format: ${VIRTUAL_ENV}/scripts/black ${VIRTUAL_ENV}/scripts/isort
-	isort --atomic terraso_backend
-	black terraso_backend
+format: ${VIRTUAL_ENV}/scripts/ruff
+	ruff format terraso_backend
 
 install:
 	pip install -r requirements.txt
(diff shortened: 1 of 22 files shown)


terraso_backend/tests/graphql/test_sites_soil_data_query.py
```diff
@@ -25,8 +25,8 @@
 
 def test_query_site_soil_data_fields(client, user):
     site = mixer.blend(Site, owner=user, name="name")
-    SoilData.objects.create(site=site, bedrock=1),
-    SoilMetadata.objects.create(site=site, selected_soil_id="a"),
+    (SoilData.objects.create(site=site, bedrock=1),)
+    (SoilMetadata.objects.create(site=site, selected_soil_id="a"),)
```
knipec, line 29: I know this isn't your code, but I believe the trailing commas and parentheses could be removed here, since they're just creating tuples that don't get used in any way

Makefile
knipec, line 1: Same question as the `uv` PR: Does this have any implications for how we'd use anything in this repo? I assume we'll just use the same make commands and the same input/outputs, but is there anything else to be aware of?

knipec (approved):
Approving assuming this change doesn't affect how we use anything. If there are any impacts we should know about as devs, please let me know!

pyproject.toml
```diff
@@ -1,6 +1,15 @@
-[tool.black]
+[project]
```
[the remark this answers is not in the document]
paulschreiber: Yes. We had this in soil-id-algorithm, but not here, so I added it.

terraso_backend/tests/graphql/test_sites_soil_data_query.py
paulschreiber: `ruff` put them there. It's possible there is a formatter rule to remove them — happy to adjust that if you'd like.

Makefile
paulschreiber: No changes to workflow. Same `make format` and `make lint`.

terraso_backend/tests/graphql/test_sites_soil_data_query.py
knipec: I suspect `ruff` only added the parentheses because the commas were already there (unnecessarily, I believe)

PR #1696, opened 2025-06-13
fix: Docker should build the db container on arm64 architecture

knipec:
## Description
I started getting an error ‘no matching manifest for linux/arm64/v8 in the manifest list entries’ when running ‘make run’. 
My understanding is that the postgis/postgis image we use only specifies that it can run on linux/amd64, but my mac uses Apple Silicon’s ARM architecture. Docker has a setting to use amd64 emulation on Apple Silicon, but it seems like this change was needed to get Docker to actually build the container from the image.

I’m not sure why it was working before, though…? It started happening after a git pull and after deleting a bunch of docker objects in Docker Desktop to free up disk space.

### Play-by-play
Made the change per this: https://stackoverflow.com/questions/65456814/docker-apple-silicon-m1-preview-mysql-no-matching-manifest-for-linux-arm64-v8 
And confirmed that the listed postgis OS/Arch does not include ARM64: https://hub.docker.com/r/postgis/postgis/tags
Backend is now running: 
Screenshot 2025-06-12 at 10 46 08 PM

### Verification steps
Do `make run` on a macbook with Apple Silicon

### DIFF
--- a/docker-compose.base.yml
+++ b/docker-compose.base.yml
@@ -14,6 +14,8 @@ services:
     tty: true
 
   db:
+    # We can remove the 'platform' line if postgis makes an official multi-platform build available
+    # See https://hub.docker.com/r/postgis/postgis/tags
     platform: linux/amd64
     image: postgis/postgis:16-master
     labels:
--- a/docker-compose.base.yml
+++ b/docker-compose.base.yml
@@ -14,6 +14,7 @@ services:
     tty: true
 
   db:
+    platform: linux/amd64
     image: postgis/postgis:16-master
     labels:
       org.techmatters.project: terraso_backend
--- a/scripts/rebuild.sh
+++ b/scripts/rebuild.sh
@@ -3,7 +3,7 @@ set -e
 
 build_hash()
 {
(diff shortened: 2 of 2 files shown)


paulschreiber:
@knipec What happens to folks running Docker on x86 platforms? Will this still work?

johannesparty:
Are you building the image with "--platform linux/amd64,linux/arm64".  It sounds like there is my arm64 linux version of some components, so probably should just use linux/amd64.  And if you do that, do you need the platform line in the Dockerfile?

garobrik (approved):
LGTM! i'd suggest leaving a comment explaining that this line can be removed if/when `postgis` have an official multi-platform build available

knipec:
Thanks @robert-bo-davis !

One thing I still don't understand is: why it would have been working in the past? On the same Apple Silicon machine, with the same repo, I had been successfully running everything for over a year, presumably via the emulation enabled in Docker. And one day it started giving me this arm64 error. I suppose something could have changed with Docker (?) but I haven't figured that part out.

PR #1757, opened 2025-07-25
fix: Show memberships for story map owner

josebui:
## Description

- Added fix to show memberhips for stroy map owners

### Checklist
- [x] New tests added

### Related Issues
- https://github.com/techmatters/terraso-web-client/issues/1632

### DIFF
--- a/terraso_backend/apps/collaboration/graphql/memberships.py
+++ b/terraso_backend/apps/collaboration/graphql/memberships.py
@@ -93,6 +93,10 @@ class MembershipListNodeMixin:
             if len(self.account_memberships):
                 return self.memberships
             return self.memberships.none()
+        can_view_memberships = getattr(self, '_can_view_memberships', False)
+        if can_view_memberships:
+            return self.memberships
+
         user = info.context.user
         is_member = self.is_member(user)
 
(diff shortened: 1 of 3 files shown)


garobrik (approved):
LGTM!

question: is there a reason to have the owner not be a member? i think groups and landscapes also have a `created_by`, why don't they need to have the same logic?

josebui:
@garobrik
I think they should all have the same logic, but I don't think there is anything defined globally on how to handle the owners of the parent entity, so I think it will require a significant level of effort to define that right now. Since nothing has been reported for the other owner entities my suspicion is that they are working as expected, so every owner logic is beinig handled explcitely for each case.

I am also worried about the potential implications for this if we devcde to make the owner jut an extra member, since it is a special type of member and we might again go back to the same issue, having to go back to the owner entity to verify if the specific user is the owner and act according to that. So I think if we have a explicit owner membeship added that might solve the problem, but I don't know if that logic holds for all cases.

Maybe as a summary, I think we need to think about this more to define a proper solution that can solve the ownership member issue in general. I do feel that this as a temporary fix works right now until we define a more complete solution for all collaboration entities.

Let me knoww what do ytou think

PR #1786, opened 2025-08-12
build: update python to 3.13.6

paulschreiber:
## Description
update python to 3.13.6


knipec (approved):
Question out of curiosity: Dependabot didn't prompt this, right? It's just that you keep up on what python updates there are?

paulschreiber:
@knipec Yes — we have to check manually. Dependabot doesn't check language versions, just module versions.

Issue #1836, opened 2025-09-18
[Rate Soil] Update server data model


knipec:
## Acceptance criteria 
- I think the main QA for this task is that we didn't break behavior with the feature flag off

## Dev notes
- SoilMetadata stores a map of values from the **soil match name** to the **user's soil rating/appraisal** (Yes / No / Unsure)
- Keep the `selected_soil_id` to maintain behavior with feature flag off, and to repurpose for custom soil name entry

PR #1847, opened 2025-09-24
feat: add user_match_ratings to SoilMetadata

knipec:
## Description
- [Draft as of 2025-09-24] Add a backend model to store soil match and user rating
- Also ruff prevents me from committing unless I update the precommit yaml?

### Related Issues
Implements https://github.com/techmatters/terraso-backend/issues/1836


terraso_backend/apps/soil_id/migrations/0023_usermatchrating.py
```diff
@@ -0,0 +1,57 @@
...
+    UserMatchRating = apps.get_model("soil_id", "UserMatchRating")
+
+    for soil_metadata in SoilMetadata.objects.all():
+        old_selected_soil_id = soil_metadata.selected_soil_id
+        if (old_selected_soil_id):
+            UserMatchRating.objects.create(
+                soil_metadata=soil_metadata,
+                match_id=old_selected_soil_id,
+                user_rating="SELECTED"
+            )
```
garobrik, line 21: hrmph: since this is running at migration time instead of when the feature flag gets flipped on for prod users, it means the initial state of people's `UserMatchRatings` will reflect what soil they had selected when the migration ran. but we don't have control over when an individual users' feature flag gets flipped, it's just whenever their app updates 🤔

Issue #1695, opened 2025-06-11
LandPKS app displays the 8th -12th location ranking returns in some scenarios


DerekCaelin:
Backend returns 8th -12th location ranking returns 

## Steps to reproduce

1. Create At location `41.66272, -72.68237'
2. See initial return of soils
<img src="https://github.com/user-attachments/assets/5e7d2b57-4c1a-49b9-bb35-80363278b23a" width="150px">

3. Create custom soils and add depth, 0-23cm
4. Add texture "Sandy Loam" (this matches "Manchester" profile)
5. Review soil ID
6. See issue

## Expected results
See 12 results, with Manchester elevated

## Actual results
See 5 results, not including Mancester
<img src="https://github.com/user-attachments/assets/92f35ade-2197-4fa3-9dd1-0893355a9cac" width="150px">

Initial rank_soils return:

```
{
   "metadata":{
      "location":"us",
      "model":"v2"
   },
   "soilRank":[
      {
         "name":"Udorthents3",
         "component":"Udorthents",
         "componentID":25664039,
         "score_data_loc":0.499,
         "rank_data_loc":"1",
         "score_data":0.499,
         "rank_data":"1",
         "score_loc":0.5,
         "rank_loc":"Not Displayed",
         "componentData":"Site data only"
      },
      {
         "name":"Branford3",
         "component":"Branford",
         "componentID":25664315,
         "score_data_loc":0.469,
         "rank_data_loc":"2",
         "score_data":0.499,
         "rank_data":"2",
         "score_loc":0.439,
         "rank_loc":"Not Displayed",
         "componentData":"Site data only"
      },
      {
         "name":"Wethersfield2",
         "component":"Wethersfield",
         "componentID":25664856,
         "score_data_loc":0.438,
         "rank_data_loc":"3",
         "score_data":0.5,
         "rank_data":"3",
         "score_loc":0.377,
         "rank_loc":"Not Displayed",
         "componentData":"Site data only"
      },
      {
         "name":"Cheshire2",
         "component":"Cheshire",
         "componentID":25664854,
         "score_data_loc":0.375,
         "rank_data_loc":"4",
         "score_data":0.5,
         "rank_data":"4",
         "score_loc":0.251,
         "rank_loc":"Not Displayed",
         "componentData":"Site data only"
      },
      {
         "name":"Wilbraham3",
         "component":"Wilbraham",
         "componentID":25664421,
         "score_data_loc":0.354,
         "rank_data_loc":"5",
[...]

PR #1848, opened 2025-09-25
fix: return all soil components even when component changes within group

garobrik:
## Description
Take advantage of the new information returned by the algorithm in https://github.com/techmatters/soil-id-algorithm/pull/331 to ensure that we always return a soil component for each group, even if the specific component changes after the user has input data.

### Checklist
- [ ] New tests added

### Related Issues
Resolves, when QA'd, #1695 

### Verification steps
Should no longer be able to repro #1695


johannesparty (approved):
I read the whole thing and see what you are doing and didn't see anything that worried me.  With that said, my review was somewhat cursory -- if you want a more thorough review, please let me know.

I did notice in your description, you did not check "new tests added", but actually you did add new tests because you included a bunch where only location is specified.

PR #1867, opened 2025-10-02
feat: SoilMetadata's selected_soil_id now derives from new user_ratings set

knipec:
## Description
- Add a `user_ratings` to the SoilMetadata django model as a database table column
- Remove the `selected_soil_id` column from the database, but now it's a computed property on the model that is derived from data in `user_ratings`. This is so the backend will still support old clients that don't have the feature flag enabled yet.
- To support this change
  - Make a migration 
  - Fix up the mutation
- FYI was trying out Claude Code for this, so much was generated by Claude, but reviewed by me. 
- Also the linter wouldn't let me commit until I formatted this yaml file, so there's that too 🤷 

Still to do: 
- The part in the mutation for `user_ratings` may get deleted to support an offline solution
- Nothing works yet with the feature flag on

### Checklist
- [x] New tests added - tests updated

### Related Issues
Implements https://github.com/techmatters/terraso-backend/issues/1836

### Verification steps
- Run the client with the FF_select_soil feature flag off, make sure the "This is the correct soil" checkbox works as expected
- Theoretically, do a GraphQL query for userSoilData with the soilMetadata and make sure it returns stuff. (But I'm not sure if this requires authentication, @johannesparty maybe we could chat since I know you've thought about this?)

### DIFF
--- a/terraso_backend/apps/soil_id/graphql/soil_metadata/mutations.py
+++ b/terraso_backend/apps/soil_id/graphql/soil_metadata/mutations.py
@@ -15,6 +15,7 @@
 
 import graphene
 from django.db import transaction
+from django.forms import ValidationError
 
 from apps.graphql.schema.commons import BaseWriteMutation
 from apps.graphql.schema.constants import MutationTypes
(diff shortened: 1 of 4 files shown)


terraso_backend/apps/graphql/schema/schema.graphql
johannesparty, line 1: "type UserDataEntry" and "input userRatingInput" seem redundant, but maybe good in case we want them to be different.  also one is type and one is input which might better (I don't know!)

terraso_backend/apps/soil_id/graphql/soil_metadata/mutations.py
johannesparty, line 1: curious about using kwargs.pop("xyz") vs kwargs.get("xyz").

johannesparty (approved):
Looks good, but I did have some questions.

terraso_backend/apps/graphql/schema/schema.graphql
knipec: 1. From chatting with garo: The redundant types are necessary because django requires types you include in mutation arguments to be "input" but types you include in query arguments to be "type".
2. Old query/mutation aren't marked deprecated yet; I expect to do that in a later task
3. Thanks for these thoughts -- I was thinking that: 
  - Offline functionality works by making changes to the client-side state, then when you go back online it’ll send a diff of the old state and the new state. ([More details here](https://github.com/techmatters/terraso-mobile-client/blob/main/dev-client/docs/Offline.md#a-connectivity).) TBD but I expect we will send a set of ratings, though it may not be the full list. 
  - If the list of soils changes: I imagine we may still keep existing entries, but entries for soils not displayed will just be ignored. I imagine the lists won’t change often. We could probably ensure this doesn’t happen, but I don’t see much benefit so far.
  - Yep, I also assume none of the existing ratings are deleted even if soil no longer exists, but new soils would return UNSURE.
  - Oh interesting case that I didn't fully understand when we chatted earlier! I’m thinking when a soil is selected, we can make sure that all soils in the *ratings* – not just soils in the displayed matches – are removed or updated to UNSURE or REJECTED (TBD which). 
  - I’m confirming behavior with Courtney, but if a soil is selected I believe we want all the other soil tiles to turn grey. I think perhaps that should be done purely on the frontend.

PR #1900, opened 2025-10-22
feat: Make new big push mutation and deprecate old one

knipec:
## Description
- Make a "big push" mutation that allows the client to push soilData and/or soilMetadata. The updates to soilMetadata mirror the existing soilData objects very closely. (I felt the duplication was fine, but feel free to disagree.) There is now a top-level error field for each data slice, in addition to the per-site error fields.
- Name the new "big push" `pushSiteData`. (Note: the naming is somewhat inconsistent throughout the repos -- example: client names a query `userSoilData` but it grabs all site and project data. I'm open to other names here.)
- Deprecate `pushSoilData` (I chose to make a new mutation instead of extend this one to be able to separately name errors and results by data slice)
- Written significantly by Claude

### Related Issues
To address https://github.com/techmatters/terraso-mobile-client/issues/3107

### Verification steps
Run tests for now
(After client-side work, the client should be able to successfully update the database with the new mutation.)

### DIFF
--- a/terraso_backend/apps/soil_id/graphql/soil_metadata/push_mutation.py
+++ b/terraso_backend/apps/soil_id/graphql/soil_metadata/push_mutation.py
@@ -59,6 +59,10 @@ class SoilMetadataPushInputEntry(graphene.InputObjectType):
 
 
 class SoilMetadataPush(BaseWriteMutation):
+    """
+    Note: as of 2025-10 this mutation is not exposed on its own; it is only used as part of SiteDataPush (similar to how SiteDataPush uses SoilDataPush).
+    """
+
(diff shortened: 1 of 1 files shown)


terraso_backend/apps/soil_id/graphql/site_data/push_mutation.py
```diff
@@ -0,0 +1,100 @@
...
+        soil_metadata_results = None
+        soil_metadata_error = None
+
+        if soil_data_entries:
+            try:
+                result = SoilDataPush.mutate_and_get_payload(
+                    root, info, soil_data_entries=soil_data_entries
+                )
+                soil_data_results = result.results
+            except Exception as e:
```
knipec: What errors will client see in various cases? Does it need to be a future backlog item?

PR #1954, opened 2025-12-02
fix: Pin GDAL to 3.11.3 using Debian snapshot archive

johannesparty:
Replace rolling Debian testing repo with frozen snapshot archive to ensure system GDAL version matches Python gdal==3.11.3 bindings.

Changes:
- Use Debian snapshot (20250822) with GDAL 3.11.3+dfsg-1
- Pin GDAL packages to version 3.11.3* with priority 1000
- Add explicit gdal==3.11.3 to requirements/base.in

This fixes version mismatch issues and provides reproducible builds by preventing unexpected GDAL version updates.

This is the simplest fix, but probably the right one for now.
We could to go 3.11.4 pretty easily ; 3.11.5 missing some bindings.
We could also go to 3.12, but that would require more testing.

NOTE: the PR double purpose is to make sure the GitHub build works; assuming it does, the PR is probably good.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,11 +7,10 @@ ENV PATH=/home/terraso/.local/bin:$PATH
 # see https://github.com/aws/aws-cli/tags for list of versions
 ENV AWS_CLI_VERSION=2.8.12
 
-# Add testing sources and pin the GDAL packages to testing
-# Allows us to get 3.10.x versions of GDAL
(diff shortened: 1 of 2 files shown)


garobrik (approved):
LGTM! might want to also create a PR pinning `gdal` in `soil-id-algorithm` in the same way, since they'll need to be kept in sync. not sure why the lint job is failing 🤔

PR #1955, opened 2025-12-02
fix: Add drift migrations for model changes since 2022

johannesparty:
Captures model changes that were applied to production but never recorded in migrations:

- story_map/0007: Add validate_name validator to StoryMap.title
- shared_data/0019: Add db_index to deleted_at fields, update ForeignKey
- core/0052: Multiple changes from django-safedelete 1.3.0 upgrade
- project_management/0029: Update ProjectSettings Meta options

All migrations use SeparateDatabaseAndState to update Django's migration state without changing the database (production already has these changes).

IMPORTANT: These migrations are irreversible to prevent accidental production breakage.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,12 +12,12 @@ repos:
         pass_filenames: false
 
 -   repo: https://gitlab.com/vojko.pribudic.foss/pre-commit-update
-    rev: v0.8.0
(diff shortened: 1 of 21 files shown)


garobrik (approved):
looks ready to me! much thanks for taking this on

PR #1959, opened 2025-12-06
feat: export (clean)

johannesparty:
## Description
Backend changes to implement export / share functions.

Organized into 5 commits:
- add export token storage and GraphQL API - models, migrations, types, queries, mutations, signals, schema
- add HTTP export endpoints for CSV/JSON downloads - urls, views, html_pages, fetch_*, transformers, formatters, static
- add is_system_export bypass for export endpoints - sites.py, projects.py, memberships.py
- add export app infrastructure - settings.py, init.py, apps.py 
- add development scripts and documentation - scripts, .env.sample, .gitignore, CLAUDE.md, handlers.py, schema.graphql

### DIFF
--- a/terraso_backend/apps/export/handlers.py
+++ b/terraso_backend/apps/export/handlers.py
@@ -13,6 +13,30 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see https://www.gnu.org/licenses/.
 
+"""
+CORS handler for export URLs.
+
+When is CORS needed?
+--------------------
+CORS is ONLY needed when JavaScript running in a browser makes cross-origin
+requests (fetch/XMLHttpRequest to a different domain).
+
+CORS is needed for:
+  - A website using JavaScript to fetch from /export/token/* URLs
+
+CORS is NOT needed for:
+  - Clicking a link to download CSV/JSON (direct browser navigation)
+  - Mobile apps making HTTP requests
+  - curl, wget, or any server-side code
+  - The HTML landing pages (same origin as API)
+
+Security note:
+--------------
+Enabling CORS for /export/token/* is low risk because the security model
+is token-based: anyone with a valid token can access the data regardless
+of CORS. The token (a 128-bit UUID) IS the security, not CORS.
+"""
+
 from corsheaders.signals import check_request_enabled
 
 
@@ -26,9 +50,6 @@ def cors_allow_export_urls(sender, request, **kwargs):
     ID-based exports (/export/id/*) require JWT authentication and follow
     the standard CORS policy (CORS_ORIGIN_WHITELIST), so they return False here
     to let the default CORS middleware behavior apply.
-
-    This handler is called by django-cors-headers middleware to determine
-    if CORS should be enabled for a particular request.
 
     Returns:
         bool: True if CORS should be enabled (for /export/token/* URLs), False otherwise


terraso_backend/apps/core/migrations/0054_create_system_export_user.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def create_system_export_user(apps, schema_editor):
+    """
+    Create the system export user used by export endpoints.
+    This user bypasses user-based security filtering to allow exports.
+    """
+    User = apps.get_model('core', 'User')
+    User.objects.get_or_create(
+        email='system-export@terraso.org',
```
garobrik: question: is this a real email?

scripts/database/restore_from_render.sh
garobrik: i feel a bit scared about this file. it's simultaneously:
- untested
- meant to be passed in a live database URL (i assume not meant to be production, but there's no safeguards around this)
- a long bash script
- contains commands like "DROP DATABASE"

i wasn't able to see any issues with it, but i think it'd feel a bit less scary if the part that interacts with the live database was isolated from everything else. i.e. there was a short, clear, uncomplicated script which runs pg_dump against the database, and then any other necessary logic happens in a different script which doesn't have access to DB credentials.

terraso_backend/apps/export/graphql/types.py
```diff
@@ -0,0 +1,31 @@
...
+import graphene
+from graphene_django import DjangoObjectType
+
+from ..models import ExportToken as ExportTokenModel
+
+
+class ResourceTypeEnum(graphene.Enum):
+    USER = "USER"
+    PROJECT = "PROJECT"
+    SITE = "SITE"
```
garobrik: question: can this also be removed now that only users can own export tokens?

terraso_backend/apps/graphql/schema/schema.py
```diff
@@ -224,6 +229,8 @@ class Mutations(graphene.ObjectType):
     save_group_membership = GroupMembershipSaveMutation.Field()
     delete_group_membership = GroupMembershipDeleteMutation.Field()
     update_shared_resource = SharedResourceUpdateMutation.Field()
+    create_export_token = CreateExportToken.Field()
+    delete_export_token = DeleteExportToken.Field()
```
garobrik: question: the naming conventions and method for adding the query endpoints to the schema differ from all the other apps. is there any reason to not align them?

terraso_backend/apps/export/handlers.py
```diff
@@ -0,0 +1,40 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see https://www.gnu.org/licenses/.
+
+from corsheaders.signals import check_request_enabled
+
+
+def cors_allow_export_urls(sender, request, **kwargs):
```
garobrik: question: how are these URLs intended to be called? i thought CORS was only needed so that JavaScript loaded from some other domain could make a request to an endpoint on your domain, but that it wasn't needed if the user is just opening the URL directly, but maybe i'm misunderstanding

terraso_backend/apps/export/signals.py
```diff
@@ -0,0 +1,110 @@
...
+def delete_site_export_tokens_on_soft_delete(sender, instance, **kwargs):
+    """
+    Delete export tokens when site is soft-deleted.
+    SafeDeleteModel doesn't trigger post_delete, so we use post_save
+    and check if deleted_at is set.
+
+    This deletes all tokens for this site (from any user).
+    """
+    if instance.deleted_at is not None:
+        # Delete all tokens for this site (any user)
```
garobrik: question: can this code also be removed since only users can own tokens now?

CLAUDE.md
```diff
@@ -0,0 +1,198 @@
...
+- `make lock-dev` - Lock development dependencies
+- `make install` - Install production dependencies
+- `make install-dev` - Install development dependencies
+
+### Database Management
+- `APP_MIGRATION_NAME="app_name migration_number" make migrate` - Run specific migration
+- `APP_MIGRATION_NAME="app_name migration_name" make print_migration_sql` - View migration SQL
+- `make makemigrations` - Create new migrations
+- `make showmigrations` - Show migration status
+- `./scripts/database/restore_from_render.sh` - Restore database from Render backup
```
garobrik: the implication of this being in CLAUDE.md seems to be that claude code would be operating in an environment where it has access to live database URLs, which i hope is not the case!

terraso_backend/apps/export/views.py
```diff
@@ -0,0 +1,306 @@
...
+    """Set up request with system export user and flag."""
+    User = get_user_model()
+    try:
+        service_user = User.objects.get(email="system-export@terraso.org")
+    except User.DoesNotExist:
+        raise RuntimeError(
+            "System export user not found. Run migrations to create 'system-export@terraso.org' user."
+        )
+    request.user = service_user
+    request.is_system_export = True
```
garobrik: still trying to wrap my head around why we need the system export user: it seems that a user should not be able to create an export token for resources to which they do not have access. so if the resources associated with the export token are accessible by the user, would we be able to set the request's user to the user who owns the export token, and then rely on our existing authentication logic rather than special-casing with the `system-export` user? or is there some reason i'm not thinking of why this wouldn't work?

terraso_backend/apps/export/graphql/mutations.py
```diff
@@ -0,0 +1,155 @@
...
+        # Users can only create tokens for themselves
+        return str(user.id) == resource_id
+
+    elif resource_type == "PROJECT":
+        # Any project member can create their own token
+        try:
+            project = Project.objects.get(pk=resource_id)
+            return project.membership_list.memberships.filter(
+                user=user,
+                deleted_at__isnull=True,
```
garobrik: question: did you try without the `deleted_at__isnull` check and it didn't work? i thought the soft delete functionality handles this stuff automatically, but i know there are other places in the app where we use a check like this, and i've never understood when/why it is necessary.

terraso_backend/apps/export/graphql/mutations.py
```diff
@@ -0,0 +1,155 @@
...
+                ).exists()
+
+            return False
+        except Site.DoesNotExist:
+            return False
+
+    return False
+
+
+def user_owns_token(user, token_obj):
```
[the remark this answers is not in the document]
johannesparty: ok being consistent and following convention is a good idea (usually).

I made the change in commit "refactor: use django-rules for export token permissions"
I kind of liked the single function to check before, but this is more consistent and fine.

terraso_backend/apps/graphql/schema/schema.py
johannesparty: no reason.  sounds like you'd like:

- CreateExportToken -> ExportTokenAddMutation
- DeleteExportToken -> ExportTokenDeleteMutation
- create_export_token -> add_export_token

changing this will require changing the client too, which is mildly painful.
on the other hand -- much easier now than later -- and it does look a lot better :)

done, commit: "refactor: rename export token mutations to match codebase conventions"
(also made change to client, so I'll need to make sure to coordinate deployment of server before releasing new client).

terraso_backend/apps/export/views.py
garobrik: i guess my opinion would be that it's cleaner & more aligned with the actual intention to be doing something more like an assume role here, since otherwise we're kind of duplicating the permissions logic (which is quite simple in this case, so not a big deal. but i also see how that could seem riskier in other ways, since the system export user only gets to bypass permissions checks in specific blessed ways, whereas an assume role could create other holes.

maybe we can have a quick convo about this? it does seem like it'd simplify the code

CLAUDE.md
garobrik: hmm okie, let's get on the same page about this!

terraso_backend/apps/graphql/schema/schema.py
garobrik: nice! the queries are also fairly different:
- they're included in the main schema via inheritance, which none of the other queries are
- they don't match the `Node` suffix
- they return an array directly rather than a `Connection` class

i actually wouldn't mind if all of our queries were defined like that, but i do think consistency is worthwhile when trying to build a codebase for the long haul

johannesparty:
ok thanks!  syrupy sounds interesting; i played around with better diffs
but seemed too heavy, maybe syrupy will do it without adding too much code.
i'll look into the cors stuff; tbh i'm really not sure if it is needed now.

thanks,

Johannes Schmidt
***@***.***
408-621-0091

PR #1981, opened 2025-12-23
feat: export minor changes to output

johannesparty:
- add pinned notes to top of site notes in export
- get rid of depthIntervalPreset from export (not correct in many cases)
- remove updatedAt field (misleading because it only includes site info, not stuff like depth data...)


knipec (approved):
Suggest documenting how to update tests if someone in the future updates export

PR #1995, opened 2026-01-06
fix: fix depth interval handling in exports

johannesparty:
Several critical bugs fixed due to misunderstanding about how depth intervals work.
(1) site preset is irrelevant if part of a project; the presets come entirely from project.
(2) custom depths can be added to presets (must be non-overlapped of course).
(3) depthdependentdata is not deleted when you change depth presets, but all the custom depths are deleted.

Some hints about the code changes:

fetch_data.py -- fetch the depth data from projects (simple)

views.py -- used to have format (csv or json) and raw (true or false); now mostly have output_format which is csv or 
json or raw.  lots of lines changed, but pretty simple.

transformers.py -- this is the bulk of the code.
the key concept: start with the 'base' intervals from the project (if it exists), or site otherwise, then add on the non-overlapping additional site custom intervals.  finally, merge the data when intervals match exactly, and output additional fields for preset value.

happy to go through the changes in person.


terraso_backend/apps/export/fetch_data.py
knipec: I guess, do we care only about the data entered for sites? Or do we also care about the depth intervals listed, even if no data is entered? Do we care about both? Do we distinguish in the exported file?

terraso_backend/apps/export/transformers.py
knipec: How well is this covered by unit tests?

terraso_backend/tests/export/test_export_transformation.py
```diff
@@ -157,30 +158,73 @@ def test_export_munsell_color_generation(self):
...
+            "project": None,
+            "soilData": {
+                "depthIntervalPreset": "CUSTOM",
+                "depthIntervals": [
+                    {"label": "0-5 cm", "depthInterval": {"start": 0, "end": 5}},
+                    {"label": "5-15 cm", "depthInterval": {"start": 5, "end": 15}},
+                ],
+                "depthDependentData": [
+                    {"depthInterval": {"start": 0, "end": 5}, "texture": "CLAY"},
+                    {"depthInterval": {"start": 5, "end": 15}, "texture": "LOAM"},
```
knipec: Could we also test what happens if there's an entry in `depthIntervals` that doesn't correspond to an entry in `depthDependendentData`, and vice versa? Or is it fine to just cover that in the csv test?

terraso_backend/tests/export/test_export_transformation.py
```diff
@@ -157,30 +158,73 @@ def test_export_munsell_color_generation(self):
...
+                    {"label": "15-30 cm", "depthInterval": {"start": 15, "end": 30}},
+                ],
+                "depthDependentData": [
+                    # Only first interval has data
+                    {"depthInterval": {"start": 0, "end": 5}, "texture": "CLAY"},
+                ],
+            },
+        }
+
+        process_depth_data_for_csv(site)
```
knipec: Wait does this change something in place, not return something new?

terraso_backend/tests/export/fixtures/Johanness-snapshot-test-BLM.csv
knipec: Nit: it's bugging me this and one other file have two `s`'s on `Johanness` but the other 4 files don't 😄

terraso_backend/apps/export/views.py
```diff
@@ -186,27 +186,46 @@ def _make_content_disposition(filename):
...
-    # Validate format parameter
-    if format not in ["csv", "json"]:
-        return HttpResponseBadRequest(f"Unsupported format: {format}")
 
-    # Raw mode only supports JSON
-    if raw and format != "json":
-        return HttpResponseBadRequest("Raw format only supports JSON output")
+    # Validate output_format parameter
+    if output_format not in ["csv", "json", "raw"]:
+        return HttpResponseBadRequest(f"Unsupported format: {output_format}")
```
knipec: Should this be part of `_validate_output_format`?

terraso_backend/tests/export/test_export_transformation.py
knipec: If there's not already, could we also have a test case where we have a project with some preset, that has a site with additional site-only depth intervals?

terraso_backend/tests/export/fixtures/Johanness-snapshot-test-BLM.csv
johannesparty: I can live with that :). If you really care I can change it.

terraso_backend/tests/export/test_export_transformation.py
johannesparty: only changing data about to be exported, not source data.

terraso_backend/apps/export/transformers.py
```diff
@@ -49,6 +53,120 @@ def _build_depth_intervals(interval_defaults):
...
+
+    Returns list of intervals in format:
+    [{"label": "0-5 cm", "depthInterval": {"start": 0, "end": 5}}, ...]
+    """
+    if preset == PRESET_NRCS:
+        return NRCS_DEPTH_INTERVALS
+    elif preset == PRESET_BLM:
+        return BLM_DEPTH_INTERVALS
+    elif preset == PRESET_CUSTOM and site:
+        # CUSTOM preset only comes from project (site CUSTOM returns None from get_effective_preset)
```
knipec: Nit: "site CUSTOM returns NONE" instead of "None"

terraso_backend/apps/export/transformers.py
```diff
@@ -295,71 +413,134 @@ def apply_object_transformations(data):
...
+
+        # Flatten depthInterval to depthIntervalStart/End
+        di = item.pop("depthInterval", {})
+        start = di.get("start")
+        end = di.get("end")
+        item["depthIntervalStart"] = start
+        item["depthIntervalEnd"] = end
+
+        # Look up interval metadata by (start, end)
+        key = (start, end)
```
knipec: Nit: Why not use `depth_key(di)`?

terraso_backend/apps/export/transformers.py
```diff
@@ -295,71 +413,134 @@ def apply_object_transformations(data):
...
+        matching_interval = interval_by_key.get(key, {})
+
+        # Use custom label if available, otherwise generate from depth range
+        if matching_interval.get("label"):
+            item["label"] = matching_interval["label"]
+        elif start is not None and end is not None:
+            item["label"] = f"{start}-{end} cm"
```
knipec: Just musing: Ah, bummer about the duplicate logic between client & backend

terraso_backend/apps/export/transformers.py
```diff
@@ -295,71 +413,134 @@ def apply_object_transformations(data):
...
+
+    effective_preset = get_effective_preset(site)
+
+    # Get preset intervals (project CUSTOM intervals if applicable)
+    preset_intervals = get_preset_intervals(effective_preset, site)
+
+    # Get site intervals that don't overlap with preset intervals
+    site_intervals = soil_data.get("depthIntervals", [])
+    site_custom = [
+        i for i in site_intervals if not any(depth_key(i) == depth_key(c) for c in preset_intervals)
```
knipec: Just musing: Ha, this is so Python

terraso_backend/apps/export/transformers.py
```diff
@@ -295,71 +413,134 @@ def apply_object_transformations(data):
...
+    # Site intervals may override preset intervals with same key
+    for interval in site_intervals:
+        interval_by_key[depth_key(interval)] = interval
+
+    # Process each measurement: flatten, add metadata, and add _depthSource
+    processed = []
+    for m in measurements:
+        item = m.copy()
+
+        # Flatten depthInterval to depthIntervalStart/End
```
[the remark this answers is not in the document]
knipec: Same question about the comment in CSV logic

knipec (approved):
Product code looks good to me! Just minor comments.

Didn't look closely through the tests because I assume we'll have a review with further test improvements soon

terraso_backend/apps/export/transformers.py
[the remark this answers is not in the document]
johannesparty: not sure what the question is.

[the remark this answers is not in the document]
knipec: Ah, never mind, I understand what this is referring to now.
We'd removed the `depth_key_flat` function because it wasn't actually used. So I wasn't sure if we were still dealing with "flattened" intervals -- but upon revisiting I see it means to "flatten" the `depthInterval` object's start and end into the `item`.

knipec (approved):
JSON "depthDependentData" could be "visibleDepthIntervals" or "_depthIntervals" since it's 1:1 with depthIntervals but not depthDependentData

johannesparty:
merging now, but won't deploy until Derek tests.

PR #2010, opened 2026-01-13
feat: add export documentation endpoint for WordPress embedding

johannesparty:
Add Django view to serve export documentation files (JS, CSS, CSV schemas) at /export/docs/<filename> with CORS enabled for cross-origin access.

This allows the LandPKS WordPress site to embed dynamic export format documentation that loads data from the API.

The code part is enabling endpoint <server>/export/docs/... with some csv, js and css files.


knipec (approved):
Also didn't review the claude-generated javascript, but talked the approach through with Johannes and sounds reasonable to me!

PR #2024, opened 2026-01-28
feat: (dev-only) console logs include graphql query/mutation info

knipec:
## Description
- The new `GRAPHQL_LOG_LEVEL` in settings.py has 3 verbosity options:
  - `minimal`: what's logged today
  - `names` (default): includes above, plus name of graphql operation and if it errored
  - `full`: includes above, plus input variables and output data
- django_structlog is middleware that handles logging. It sends django [signals](https://django-structlog.readthedocs.io/en/latest/api_documentation.html#django_structlog.signals.bind_extra_request_failed_metadata) that allow our new receivers to customize log metadata. The receivers need to be registered by importing the module.
- Note: It's annoying that logs are duplicated, but this was the case before my changes.

**Level of scrutiny**: 2/5 -- I used claude code to write this, and reviewed at a high level, but didn't read every line. I manually tested and got claude to write unit tests that I also reviewed at a high level. It should just affect logging in development environments, so seems low risk. But let me know if you think this merits closer inspection.

### Example:
(With "full" log level)
Screenshot 2026-01-28 at 11 30 30 AM

### Related Issues
Implements https://github.com/techmatters/terraso-product/issues/1514

### Verification steps
- As a developer, run the server. Take actions, and confirm you can see the names of the graphql operations in the console logs. 
- Update GRAPHQL_LOG_LEVEL in settings.py to "full" or "minimal". Make sure you see the correct amount of info in console logs.
- Run tests

### DIFF
--- a/terraso_backend/config/graphql_log.py
+++ b/terraso_backend/config/graphql_log.py
@@ -1,0 +1,158 @@
+# Copyright © 2021-2025 Technology Matters
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see https://www.gnu.org/licenses/.
+
+import json
+import re
+
+from django.conf import settings
+from django.dispatch import receiver
+from django_structlog import signals
+
+
+def _get_log_level():
+    """Read GRAPHQL_LOG_LEVEL at call time so code changes take effect on hot reload."""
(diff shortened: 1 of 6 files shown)


terraso_backend/config/graphql_log.py
```diff
@@ -0,0 +1,158 @@
...
+
+import json
+import re
+
+from django.conf import settings
+from django.dispatch import receiver
+from django_structlog import signals
+
+
+def _get_log_level():
```
johannesparty, line 24: is there documentation somewhere showing what GRAPHQL_LOG_LEVEL values might be?
I would like toes this in  .env.sample

johannesparty: ok I see settings.py describes the value.
so maybe put reference here saying "see settings.py".
but I would also add to .env.sample.

johannesparty (approved):
Looks very useful!

I am happy that you check the config for "minimal" at the top of the extraction stuff which means even if this is buggy somehow and crashes, that would only happen if higher logging turned on.

Probably not likely -- I see lots of try/catch so this should be safe.

terraso_backend/config/graphql_log.py
knipec: Ah thanks! I've moved the comments around, but more importantly am now getting the value from the .env instead of hard-coding it in settings.py

PR #2033, opened 2026-02-21
fix: fix export failure and color lookup related to soil id

johannesparty:
First commit: depth dependent data includes non-visible depths when soil.  commit has details.
Details:
- move the filter related functions from transformers.py to depth_helpers.py
- use the filter in loop in fetch_data.py preparing info to send to soil id.

Second commit: use much better color conversion algorithm.  commit has details.
- was just using super inaccurate conversion before, now doing it the correct way (which is surprisingly complicated, but not that hard to use existing algorithm).

both changes pretty simple and somewhat embarrassingly wrong until now.


terraso_backend/apps/export/transformers.py
```diff
@@ -20,129 +20,17 @@
...
 
 from apps.soil_id.models.depth_dependent_soil_data import DepthDependentSoilData
-from apps.soil_id.models.project_soil_settings import (
-    BLMIntervalDefaults,
-    DepthIntervalPreset,
-    NRCSIntervalDefaults,
-)
 from apps.soil_id.models.soil_data import SoilData
 
+from .depth_helpers import (  # noqa: F401
```
knipec: `noqa: F401` makes the linter ignore warnings about unused imports, right? If so, why does this file need to import unused names?

terraso_backend/apps/export/fetch_data.py
knipec, line 1: Might be worth a comment mentioning that there is corresponding client code (as my understanding is this should mirror that code).

knipec (approved):
I didn't review that the munsell/LAB color algorithm works as it does on the client, so am just taking your word for it on that :)

terraso_backend/apps/export/fetch_data.py
knipec, line 1: Would we be able to test that this color logic correctly converts between munsell and LAB?

johannesparty: added test file generated by client code along with test case which checks.

terraso_backend/apps/export/transformers.py
johannesparty: yep, this is leftover, now removed.

terraso_backend/apps/export/fetch_data.py
johannesparty: I will this the unit test now implies this; the json test file references the mobile-client.

PR #2040, opened 2026-03-19
feat: prepare for gunicorn gthread workers to improve concurrency

johannesparty:
Currently two workers each one thread so at most two concurrent calls to backend.

Pretty bad if doing export which can take a long time.
Might be bad for offline when client coming online does a bunch of things all at once (keeping server busy).
I say "might" because not sure how much effect in the real-world (usually the server isn't very busy at all).

This change should only affect scalability, not behavior and is not required for offline, but might improve performance.

Make shared state thread-safe for gthread worker model:
- Use threading.local() for soil ID database connection (was global shared)
- Add threading.Lock() around export caches and lazy-loaded munsell table
- Set CONN_MAX_AGE=600 with health checks for persistent thread connections
- Add make run-gunicorn target for local concurrency testing

To enable in production, set: GUNICORN_CMD_ARGS=--worker-class gthread --threads 8

With WEB_CONCURRENCY=2 and GUNICORN_CMD_ARGS=--worker-class gthread --threads 8, up to 16 calls can be handled at once, which should be plenty and this shouldn't increase our server cost at all. 

If needed, both numbers could be increased, but there are some tradeoffs; more CONCURRENCY scales well but uses ~350MB for each so eventually we'd need to play for more memory.  --threads costs almost nothing and scales well if all IO bound, but not so well for CPU work.  probably we are mostly IO bound, but not entirely, some of the algorithms for soil id, etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,19 +104,14 @@ setup-git-hooks:
 run: check_rebuild
 	@./scripts/docker.sh "$(DC_FILE_ARG)"
 
-# Run with gunicorn + gthread workers to match production threading behavior.
+# Run with gunicorn to match production threading behavior.
 # Unlike "make run", this does NOT auto-reload on code changes — restart manually.
-# Useful for testing concurrency (e.g. parallel curl requests to confirm threading).
-# Production recommendation: --workers 2 --threads 8 (16 concurrent requests).
-# Workers are OS processes (~350MB each); threads are cheap and share memory.
-# Threads help most for I/O-bound work (DB queries, external API calls).
+# Reads WEB_CONCURRENCY and GUNICORN_CMD_ARGS from .env (same values used on Render).
+# Useful for testing concurrency locally before deploying to production.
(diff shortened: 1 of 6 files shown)


Makefile
knipec, line 1: Would it make sense for local to also get the settings from environment variables, if that's how we do it in staging/prod? 
If not, might be helpful to note what the corresponding env keys are for production threading behavior?

knipec (approved):

Makefile
johannesparty: good idea, I changed this so "make run-gunicorn" also just uses the values from .env.

you should probably add "GUNICORN_CMD_ARGS=--worker-class gthread --threads 8" to your .env file (it has been added to .env.sample).

the README.md now explains when to use run-gunicorn and when not to use it:

image

knipec (approved):
Thanks! I find the updated changes much more understandable for how to use it as a developer.

PR #2041, opened 2026-03-20
fix: prevent sensitive settings from leaking in error pages

johannesparty:
Not really sure I like the DEBUG=False change, but the DATABASE_URL change for error reports seems important.
Let me know what you think about the DEBUG=False forced in prod.

- Force DEBUG=False in non-development environments regardless of env var
- Add DATABASE_URL to Django's hidden_settings pattern so database passwords are redacted in error reports


terraso_backend/config/settings.py
```diff
@@ -37,6 +39,14 @@
 ENV = config("ENV", default="development")
 
 DEBUG = config("DEBUG", default=False, cast=config.boolean)
+if ENV != "development":
+    DEBUG = False
```
garobrik, line 43: comment: perhaps we still want debugging to be available on staging? i'm ok with not doing this though, since it's mainly there to protect ourselves from enabling debug on prod (if someone has access to our prod env vars we probably have bigger fish to fry). and for that use case (preventing ourselves from enabling), i'd rather fail loudly than do a silent override like this

terraso_backend/config/settings.py
```diff
@@ -37,6 +39,14 @@
...
+
+# Hide database URLs (which contain passwords) in Django error pages.
+# Django's defaults already hide settings matching API, TOKEN, KEY, SECRET, PASS.
+SafeExceptionReporterFilter.hidden_settings = re.compile(
+    r"API|TOKEN|KEY|SECRET|PASS|SIGNATURE|HTTP_COOKIE|DATABASE_URL", re.IGNORECASE
+)
```
garobrik: were you able to validate that this works as intended? django's documentation seems to imply that this is not the blessed way to customize this behavior: https://docs.djangoproject.com/en/6.0/howto/error-reporting/#custom-error-reports

garobrik (commented):
i don't have a problem with the changes conceptually! but i'm a bit less certain about the implementation details

terraso_backend/config/settings.py
johannesparty: I'm not sure, let's try it like this for now, since it is the safer choice.  if it is too much, we can remove it.

terraso_backend/config/settings.py
```diff
@@ -37,6 +39,21 @@
 ENV = config("ENV", default="development")
 
 DEBUG = config("DEBUG", default=False, cast=config.boolean)
+if ENV != "development":
+    DEBUG = False
+
+
+# Extend Django's default hidden_settings to also redact DATABASE_URL (which contains
+# passwords) in error pages. Uses subclass + DEFAULT_EXCEPTION_REPORTER_FILTER per
+# https://docs.djangoproject.com/en/6.0/howto/error-reporting/#custom-error-reports
```
johannesparty, line 48: changed to recommended implementation.

terraso_backend/config/settings.py
johannesparty: I changed this so it should do what the doc suggests.

garobrik (commented):
ok no issues with the code now! still wondering if you were able to validate whether it is working as intended, even just ad-hoc?

johannesparty:
- I tested DATABASES/DATABASE_URL  are turned to ****.

- I tested the stack frame filters out sensitive for sync_s3_buckets, backup_service, restore, and soil_id_database_connection (which was what I was trying to do, doesn't appear other places)

- I didn't test DEBUG forced off outside of development, but the code is simple and if it doesn't work ... not a huge problem.

garobrik (approved):
great, sounds good, thank you!

PR #2043, opened 2026-04-02
feat: create sites offline (backend)

knipec:
## Description
Add SitePush as a sub-mutation to the UserDataPush. SitePush can add or update sites, and can add, update, and delete site notes. 

We try to match behavior of SoilDataPush and SoilMetadataPush sub-mutations as best as possible.
As with other sub-mutations, if an entity level failure happens within the SitePush, we will not apply the changes to that entity, but will consider the site synced.

**Notes on edge cases**:
- If user creates a site with a deleted/unknown project, we will quietly convert it to an unaffiliated site.
- There is a SitePushHistory now (similar to SoilDataPushHistory), so we can see input data even if a failure happened.
- UNEXPECTED_ERROR is a new entity-level failure reason for all sub-mutations. This means that if there's an unknown backend logic error while processing an "entity" (a site within a sub-mutation), we will show a sync conflict and not apply the changes but consider them synced so won't retry.

### Related Issues
Implements https://github.com/techmatters/terraso-product/issues/1144 and https://github.com/techmatters/terraso-mobile-client/issues/2198


terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -2455,18 +2460,45 @@ input SoilDataPushInputDepthInterval {
...
 To enable offline functionality.
 Pushes at least one of the following sub-mutations:
+- sites (creates/updates sites and their notes)
 - soil data
 - soil metadata (fully replaces user ratings)
 
-Partial updates are possible, if failure happens at the level of a sub-mutation, or more granularly within the sub-mutation.
+Partial updates are possible, if failure happens at the level of a sub-mutation, or more
+granularly within the sub-mutation. Sites are processed before soil data/metadata to ensure
+new sites exist on the server before their soil data references them.
```
knipec: This comment is duplicated

terraso_backend/apps/graphql/schema/sites.py
```diff
@@ -132,18 +136,38 @@ def mutate_and_get_payload(cls, root, info, create_soil_data=True, **kwargs):
...
+                client_uuid = uuid.UUID(str(client_id))
+            except ValueError:
+                raise GraphQLValidationException(
+                    error_messages=[
+                        ErrorMessage(code="invalid", context=ErrorContext(model="Site", field="id"))
+                    ]
+                )
+            existing = Site.objects.filter(id=client_uuid).first()
+            if existing:
+                return SiteAddMutation(site=existing)
```
knipec: Does the push use this? If not, 
a) Can probably remove all the changes to SiteAddMutation, right? --> Done
b) Confirm the logic about adding to a deleted project still exists in the push --> Yup, in SitePush's `_process_site_entry`.
c) I think we can also deprecate these? Now all site edits (online and offline both) will go through this codepath. --> Done (except not deprecating `update_site` as it is still used for transferring sites, which is online-only and the only way to transfer sites currently)

terraso_backend/apps/project_management/graphql/site_notes.py
```diff
@@ -45,18 +49,40 @@ class SiteNoteAddMutation(BaseWriteMutation):
...
+                raise GraphQLValidationException(
+                    error_messages=[
+                        ErrorMessage(
+                            code="invalid", context=ErrorContext(model="SiteNote", field="id")
+                        )
+                    ]
+                )
+            existing = SiteNote.objects.filter(id=client_uuid).first()
+            if existing:
+                return SiteNoteAddMutation(site_note=existing)
```
knipec: Same -- does the push use this?

terraso_backend/apps/project_management/graphql/site_push.py
```diff
@@ -0,0 +1,311 @@
...
+    name = graphene.String()
+    latitude = graphene.Float()
+    longitude = graphene.Float()
+    elevation = graphene.Float()
+    privacy = SiteNode.privacy_enum()
+    project_id = graphene.ID()
+    # Note operations
+    new_notes = graphene.List(graphene.NonNull(SitePushNoteInput), required=True)
+    updated_notes = graphene.List(graphene.NonNull(SitePushNoteInput), required=True)
+    deleted_note_ids = graphene.List(graphene.NonNull(graphene.ID), required=True)
```
knipec, line 61: Why are these `required=True`?

terraso_backend/apps/project_management/graphql/site_push.py
```diff
@@ -0,0 +1,311 @@
...
+    site_id = graphene.ID(required=True)
+    result = graphene.Field(SitePushEntryResult, required=True)
+
+
+# NOTE: we check if the user has all permissions required to edit a site,
+#       rather than checking individual permissions against which data has been modified.
+# NOTE: we catch errors at the granularity of each site in the request.
+#       So one site's updates can succeed while another fails. But if any of
+#       an individual site's updates are invalid (including note conflicts), we
+#       reject all of that site's updates atomically.
```
knipec, line 92: Is this an issue? 
Like: if the site itself was updated successfully, but the note you try to update was deleted on the server, what happens? --> We now have a special case for this for not to be an error. Currently, the note will be deleted and we will not show the sync conflict dialog.

- I thought: The SitePushEntry entity fails, so the graphql response does not contain the site info, just the error info. The client will show the sync conflict dialog. (I feel this would be more consistent with the patterns elsewhere (like with editing a deleted site), but 🤷.

terraso_backend/apps/project_management/graphql/site_push.py
```diff
@@ -0,0 +1,311 @@
...
+        def __init__(self, reason: SitePushFailureReason):
+            self.reason = reason
+
+    @staticmethod
+    def log_site_push(user: User, site_entries: list[dict]) -> list[SitePushHistory]:
+        history_entries = []
+        for entry in site_entries:
+            site_id = entry.get("site_id")
+            site = Site.objects.filter(id=site_id).first() if site_id else None
+            history_entry = SitePushHistory(
```
knipec, line 118: Probably not worth investigating, but the SitePushHistory in the django admin panel seems not to be tied to the Site object?? 
Screenshot 2026-04-01 at 8 50 25 PM

terraso_backend/apps/project_management/models/site_push_history.py
knipec, line 1: Can we refactor to share anything with the other push history model?

terraso_backend/apps/soil_id/graphql/sync/push_mutation.py
```diff
@@ -36,12 +41,16 @@ class UserDataPush(BaseWriteMutation):
     """
     To enable offline functionality.
     Pushes at least one of the following sub-mutations:
+    - sites (creates/updates sites and their notes)
     - soil data
     - soil metadata (fully replaces user ratings)
 
-    Partial updates are possible, if failure happens at the level of a sub-mutation, or more granularly within the sub-mutation.
+    Partial updates are possible, if failure happens at the level of a sub-mutation, or more
```
knipec: "if failure happens at the level of a sub-mutation" -- I don't think this is actually a level of granularity that can currently be reflected in the response

terraso_backend/apps/soil_id/graphql/sync/push_mutation.py
```diff
@@ -50,24 +59,32 @@ class UserDataPush(BaseWriteMutation):
...
         )
 
     @classmethod
     def mutate_and_get_payload(cls, root, info, **kwargs):
+        site_entries = kwargs.get("site_entries", [])
         soil_data_entries = kwargs.get("soil_data_entries", [])
         soil_metadata_entries = kwargs.get("soil_metadata_entries", [])
 
-        if not soil_data_entries and not soil_metadata_entries:
+        if not site_entries and not soil_data_entries and not soil_metadata_entries:
```
knipec, line 75: There's gotta be some software design pattern for this

terraso_backend/apps/soil_id/graphql/sync/push_mutation.py
knipec, line 1: In general: Make sure if we make changes to this mutation, it will still support old clients.

terraso_backend/tests/graphql/mutations/test_site_notes.py
knipec, line 1: Do we still use the `addSiteNote` mutation? Is there any sense in testing this? If not, delete these tests or convert them to the actual push mutation

terraso_backend/apps/project_management/graphql/site_push.py
```diff
@@ -0,0 +1,311 @@
...
+                raise SitePush.SiteConflictError(SitePushFailureReason.NOT_ALLOWED)
+
+            SiteNote.objects.create(
+                id=note_uuid, site=site, content=note_input["content"], author=user
+            )
+
+        for note_input in site_entry.get("updated_notes", []):
+            note = SiteNote.objects.filter(id=note_input["id"]).first()
+            if note is None:
+                continue  # Idempotent: note was deleted externally, skip update
```
knipec, line 288: This means we will not show the sync conflict dialog if you try to edit a note that someone else deleted, we'll just silently get rid of your edits. Perhaps this is fine. 
But it also means we will not mess up any of the other things in the entity so that's good 🤷

terraso_backend/tests/graphql/mutations/test_site_push.py
knipec, line 1: TODO: Make sure the tests cover what you want

terraso_backend/tests/graphql/mutations/test_sites.py
knipec, line 1: Do we still use the `createSite` mutation? Is there any sense in testing this? If not, delete these tests or convert them to the actual push mutation

terraso_backend/apps/project_management/graphql/site_push.py
```diff
@@ -0,0 +1,311 @@
...
+    updated_notes = graphene.List(graphene.NonNull(SitePushNoteInput), required=True)
+    deleted_note_ids = graphene.List(graphene.NonNull(graphene.ID), required=True)
+
+
+class SitePushFailureReason(graphene.Enum):
+    SITE_DOES_NOT_EXIST = "SITE_DOES_NOT_EXIST"
+    NOTE_DOES_NOT_EXIST = "NOTE_DOES_NOT_EXIST"
+    NOT_ALLOWED = "NOT_ALLOWED"
+    INVALID_DATA = "INVALID_DATA"
+    UNEXPECTED_ERROR = "UNEXPECTED_ERROR"
```
knipec, line 68: Will this (for soilData or soilMetadata) break old clients? --> Normally it would be considered a breaking change, but I ran through a manual test and it seems fine. So probably because the client doesn't actually use the specific failureReasons anywhere, it ends up being ok to add new ones. So far...

terraso_backend/apps/project_management/graphql/site_push.py
knipec, line 1: Maybe change dialog wording to like "some of your changes couldn't be saved" (check with Courtney).

terraso_backend/apps/project_management/graphql/site_push.py
knipec, line 68: Do we log when / which of these errors happen?

terraso_backend/apps/project_management/graphql/site_push.py
knipec: Try: Just doing all of the parts and not returning the Failure early 

--> We did add the UNEXPECTED_ERROR, which will now catch unexpected backend logic errors as entity-level errors instead of failing the entire mutation. 

But for the new SitePush sub-mutation, if one of the parts of that fails (like adding/deleting/updating a note, or updating a site), it will be an entity-level error, and none of the changes for that entity will be applied (where the entity is the SitePush sub-mutation for a particular site). 
I feel feel fine about this behavior because it is analogous to how syncing soilData works. (If you edit slope shape and add a depth interval, and an exception is raised during the SoilDataPush sub-mutation, all of the soilData changes to that site will also be reverted.)

Editing a deleted note does NOT yield a failure anymore, which I don't think has an analogous case. Maybe the closest equivalent is a depth interval getting deleted by one person, but edited by another? In that case it appears we will apply the change though (but I think it's maybe a new depth interval on the backend).

knipec: Try: If the updated note was deleted, can we make a new note instead? --> Decided not to spend time on this, in call with Johannes, Courtney, Derek, Steve, Carissa on 4/7

terraso_backend/apps/graphql/schema/schema.graphql
knipec: I guess that's just an artifact of the generated api schema. It's only in `push_mutation.py` once 🤷

terraso_backend/apps/project_management/graphql/site_push.py
knipec: It kinda doesn't matter, just now the client needs to pass empty lists for them

terraso_backend/apps/project_management/graphql/site_push.py
knipec: Made a change so we will now do `posthog.capture(`${type}_push_failure`, {failures});` in the push instead of just showing the number of push failures.

(And `missing_data_sync_error` still exists in the SyncNotificationContext)

terraso_backend/apps/project_management/graphql/site_push.py
knipec: Yep. See https://github.com/techmatters/terraso-backend/pull/2043#discussion_r3030492362 for musings.

terraso_backend/apps/soil_id/graphql/sync/push_mutation.py
johannesparty: not sure it is any better but of course can do something like "if not (site_entries or soil_data_entires or soil_metadat_entries):", but not sure that is what you are looking for.

johannesparty: ok apparently another alternative is:

if not any((site_entries, soil_data_entries, soil_metadata_entries)):

this seems slightly better but IMO the original code you have is fine too.

terraso_backend/apps/project_management/graphql/site_push.py
```diff
@@ -0,0 +1,303 @@
...
+    project_id = graphene.ID()
+    # Note operations
+    new_notes = graphene.List(graphene.NonNull(SitePushNoteInput), required=True)
+    updated_notes = graphene.List(graphene.NonNull(SitePushNoteInput), required=True)
+    deleted_note_ids = graphene.List(graphene.NonNull(graphene.ID), required=True)
+
+
+class SitePushFailureReason(graphene.Enum):
+    SITE_DOES_NOT_EXIST = "SITE_DOES_NOT_EXIST"
+    NOTE_DOES_NOT_EXIST = "NOTE_DOES_NOT_EXIST"
```
knipec: Oh, we should probably delete this now? Because if it doesn't exist we'll just consider it to have been deleted with no conflict.

terraso_backend/tests/graphql/mutations/test_site_push.py
```diff
@@ -0,0 +1,697 @@
...
+
+    owned_site.refresh_from_db()
+    assert owned_site.name == "Should Succeed"
+
+
+def test_new_site_note_failure_rolls_back_site_creation(client_query, user):
+    """
+    If a note operation fails on a new site, the site itself is also rolled back.
+    (The note UUID is valid but the note already exists under a *different* site,
+     triggering a permission failure via EDIT_NOTE on the wrong site.)
```
knipec: Should we have a special case for if a newly created site or note UUID happens to overlap with an existing one, so we don't drop all the changes to the site?

terraso_backend/apps/soil_id/graphql/sync/push_mutation.py
knipec: Ah I meant this at a much higher level, not just this line of code. Since we have these 3 sub-mutations that behave similarly with respect to entity-level errors, how they are executed by the overarching push mutation, it seems like it should be a polymorphism sort of thing. This is probably more of a concern on the mobile-client than the backend, but was thinking about it here as well. Just not sure if it's worth taking time to figure out how to refactor to nicer patterns, so haven't thought very hard about it yet.

terraso_backend/apps/project_management/graphql/site_push.py
johannesparty: I think the sis vastly better than the generic sync conflict, but we'll see...

terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -1691,7 +1691,7 @@ type Mutations {
   deleteStoryMapMembership(input: StoryMapMembershipDeleteMutationInput!): StoryMapMembershipDeleteMutationPayload!
   approveStoryMapMembership(input: StoryMapMembershipApproveMutationInput!): StoryMapMembershipApproveMutationPayload!
   approveStoryMapMembershipToken(input: StoryMapMembershipApproveTokenMutationInput!): StoryMapMembershipApproveTokenMutationPayload!
-  addSite(input: SiteAddMutationInput!): SiteAddMutationPayload!
+  addSite(input: SiteAddMutationInput!): SiteAddMutationPayload! @deprecated(reason: "Use push_user_data instead.")
   updateSite(input: SiteUpdateMutationInput!): SiteUpdateMutationPayload!
```
johannesparty: I think updateSite should also be deprecated, agreed?

terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -1691,7 +1691,7 @@ type Mutations {
...
   deleteSite(input: SiteDeleteMutationInput!): SiteDeleteMutationPayload!
```
johannesparty, line 1696: Should we add offline support for deleteSite?  I'm tempted to say yes, but I"m also tempted to wait.

terraso_backend/apps/project_management/graphql/site_push.py
johannesparty: yep, just removed it.

terraso_backend/apps/project_management/graphql/site_push.py
johannesparty: decided not to do this ... if deleted, the note changes are thrown away.  think of it as saving the  note changes and deleting the note, which means changes are thrown away too.

terraso_backend/apps/graphql/schema/schema.graphql
johannesparty: ok I just deprecated it, let me know if you think that is a mistake.

johannesparty:
I didn't do a super thorough review, but I did do a fair amount of testing.
We should probably discuss before actually merging.

johannesparty (approved):

terraso_backend/apps/graphql/schema/schema.graphql
knipec: I am most tempted to wait and not scope creep! And we can touch base with @DerekCaelin and @CourtneyLee333 to see if it's worth doing after we get this feature out! :)

terraso_backend/apps/graphql/schema/schema.graphql
knipec: I think it should not be deprecated! It is still used for removing a site from project (screenshot below)
Perhaps a comment about that would have been good though :)

Screenshot 2026-04-15 at 8 27 17 PM

terraso_backend/tests/graphql/mutations/test_site_push.py
knipec: UUID generation is very unlikely to overlap, so no need

PR #2044, opened 2026-04-07
fix: backfill empty user names on subsequent sign-ins

johannesparty:
Previously _persist_user only wrote first_name/last_name when creating a brand-new user, so any user whose name was missing on first signup (notably Apple Sign In users, where Apple only returns the name on the very first authorization) was stuck with an empty name forever — even if a later sign-in via Google/Microsoft/Apple-after-revoke supplied the name.

Now, on every sign-in, if the provider sends a name and the existing user record's corresponding field is empty, we fill it in. We never overwrite a non-empty name, so users who have edited their name in the app are unaffected.

Todo: may still want to figure out why Apple doesn't return the name more often, and this might not be fixable, so we should consider adding a way for the user to edit.  But that is separate from this PR.

### DIFF
--- a/terraso_backend/tests/auth/test_token_exchange.py
+++ b/terraso_backend/tests/auth/test_token_exchange.py
@@ -179,3 +179,80 @@ def test_token_exchange_is_new_account_false(client, private_key, payload):
     contents = resp.json()
     assert contents["is_new_account"] is False
     assert User.objects.filter(email="existinguser@example.org").count() == 1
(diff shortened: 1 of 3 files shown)


terraso_backend/tests/auth/test_token_exchange.py
knipec, line 1: Nice!

knipec (approved):

PR #2048, opened 2026-05-11
feat: move .env outside repo (~/secrets/terraso-backend/.env)

johannesparty:
Please see https://sites.gitbook.com/preview/site_FIWnZ/technical-notes/privacy-and-security

Keeping the dev .env file inside the project tree means any tool browsing the repo (AI assistants, indexers, accidental scans) can read the secrets it contains. Move it out of the repo so the secrets aren't reachable just by walking the project directory.

- docker-compose.base.yml: env_file paths use ${HOME}/secrets/terraso-backend/.env (interpolated by Compose at load time).
- Makefile: ENV_FILE defaults to $(HOME)/secrets/terraso-backend/.env; passed to docker compose via --env-file so ${PORT} et al. still interpolate. Override with ENV_FILE=/some/path make ...
- .github/workflows/build.yml: 3 "set up env" steps now write to $HOME/secrets/terraso-backend/.env (matches local-dev path on ubuntu-latest runners).
- .env.sample / README.md: setup instructions point at the new location.

Migration for existing developers (one-time):
  mkdir -p ~/secrets/terraso-backend mv .env ~/secrets/terraso-backend/.env

Cloud Run / Render unaffected — they read env vars from their own service configs, not from any .env file in the repo.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,15 @@
+{
+  "permissions": {
+    "deny": [
+      "Read(~/secrets/**)",
+      "Read(.env)",
+      "Read(.env.*)",
+      "Read(.envrc)",
+      "Bash(env*)",
+      "Bash(printenv*)",
+      "Bash(cat ~/secrets/*)",
+      "Bash(cat .env*)",
+      "Bash(echo $*)"
+    ]
+  }
+}
--- a/.env.sample
+++ b/.env.sample
@@ -1,3 +1,16 @@
+# Local development env file template.
+#
+# Save your filled-in copy at ~/secrets/terraso-backend/.env (NOT inside this
+# repo). The Makefile and docker-compose.base.yml resolve env_file paths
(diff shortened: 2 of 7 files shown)


.claude/settings.json
```diff
@@ -0,0 +1,15 @@
+{
+  "permissions": {
+    "deny": [
+      "Read(~/secrets/**)",
```
knipec, line 4: Will this work on Windows machines? (Maybe not a big deal if none of our developers use windows, but if down the line people ever use Windows)

knipec (approved):

.claude/settings.json
johannesparty: not sure, but I'm guessing it'll work because they probably use WSL (windows subsystem for linux) which makes it look like linux.  but I also suspect a lot of stuff won't work under windows if not WSL -- lots or little tools and stuff we use.

PR #2051, opened 2026-06-01
fix: scope anonymous & over-broad GraphQL access (F1–F6)

johannesparty:
Stacked on #2049 — please review that one first. GitHub will auto-retarget
this PR to `main` once #2049 merges; the diff below already shows only this
branch's changes.

Closes the anonymous-access and over-broad-read findings from the security
audit. Four logical groups:

### 1. Foundation: by-id leak (`commons.py`)
`TerrasoRelayNode.get_node_from_global_id` fetched records by id directly,
bypassing each Node's `get_queryset`. So per-Node auth filters applied to
list/connection queries only — `query { site(id:) }` leaked any record by
id (F2/F3). Now routes single-id lookups through `get_queryset`. Also
returns `None` instead of raising on no match (missing vs. forbidden both
read as null).

### 2. Per-Node anonymous scoping
- **`projects.py`** (F5): anonymous fell through to `user_id=None`, which
  over the membership LEFT OUTER JOIN matched all membership-less projects.
  Now short-circuits anonymous → `.none()`.
- **`users.py` `UserNode`** (F1): anonymous → `.none()` (no user-table
  enumeration / by-id lookup).
- **`visualization_config.py`**: top-level detection switched from a
  field-name check to query-path depth, closing the singular
  `visualizationConfig(id:)` leak.

### 3. Sites: scoping + public-read feature (`sites.py`, `models/sites.py`)
Anonymous still rejected. Authenticated read now also includes
`privacy=PUBLIC` sites (`filter_visible_sites`), per product 2026-05-18 for
the future data portal. **Read-only** — writes still go through per-site
permission checks. `.distinct()` added because the membership join
multiplies rows.

### 4. Extra surface gated
- **`export/queries.py`**: explicit anonymous → `[]` (was empty only by
  coincidence).
- **`users.py` `UserAddMutation`**: `BaseAuthenticatedMutation` →
  `BaseAdminMutation` (superuser-only) (F6).
- **`users.py` `UserFilter`**: substring filters (`email__icontains`, …)
  gated to superuser to prevent user-table harvesting (F1 follow-up).

### Tests
Cross-tenant mutation probes, anonymous-scoping suite, and public-site
visibility tests added.

### DIFF
--- a/terraso_backend/apps/graphql/schema/sites.py
+++ b/terraso_backend/apps/graphql/schema/sites.py
@@ -98,7 +98,7 @@ class SiteNode(DjangoObjectType):
         user = info.context.user
         if user.is_anonymous:
             return queryset.none()
-        return sites.filter_only_sites_user_owner_or_member(user, queryset)
+        return sites.filter_visible_sites(user, queryset)
 
     @classmethod
     def privacy_enum(cls):
--- a/terraso_backend/apps/graphql/schema/users.py
+++ b/terraso_backend/apps/graphql/schema/users.py
@@ -40,6 +40,23 @@ logger = structlog.get_logger(__name__)
 
 
 class UserFilter(FilterSet):
+    # Substring filters (`email__icontains`, `first_name__icontains`,
+    # `last_name__icontains`) can be used to harvest the user table — e.g.
+    # `users(email_Icontains: "@target.com")` returns every account at a
+    # given domain.  No production client uses them: web/mobile/shared all
+    # call only `users(email)` or `users(email_Iexact: ...)` for the
+    # userProfile and add-team-member flows (verified by grep against
+    # terraso-web-client, terraso-mobile-client, terraso-client-shared).
+    #
+    # The substring filters are kept in the schema for potential future
+    # admin/support workflows but are gated to `is_superuser=True` callers
+    # in `qs` below.  Non-superuser callers using these filters get
+    # `.none()` rather than an error: an empty result keeps the schema
+    # response valid for any client that fires a substring filter
+    # accidentally, and avoids surfacing "you would need superuser to do
+    # this" as introspection-equivalent metadata.
+    ADMIN_ONLY_FILTERS = ("email__icontains", "first_name__icontains", "last_name__icontains")
+
     project = CharFilter(method="filter_user_in_project")
 
     class Meta:
@@ -53,6 +70,17 @@ class UserFilter(FilterSet):
     def filter_user_in_project(self, queryset, name, value):
         memberships = Membership.objects.filter(membership_list__project=value)
         return queryset.filter(collaboration_memberships__in=memberships)
+
+    @property
+    def qs(self):
+        # If any admin-only filter has a non-empty value, require superuser.
+        # `self.data` is the dict of incoming filter args keyed by their
+        # Django ORM lookup name (e.g. "email__icontains").
+        if any(self.data.get(name) for name in self.ADMIN_ONLY_FILTERS):
+            user = getattr(self.request, "user", None) if self.request else None
+            if not (user and user.is_superuser):
+                return super().qs.none()
+        return super().qs
 
 
 class UserNode(DjangoObjectType):
--- a/terraso_backend/apps/graphql/schema/users.py
+++ b/terraso_backend/apps/graphql/schema/users.py
@@ -28,6 +28,7 @@ from apps.core.models.users import USER_PREFS_KEY_ACCOUNT_DELETION, USER_PREFS_K
 from apps.graphql.exceptions import GraphQLNotAllowedException
 
 from .commons import (
+    BaseAdminMutation,
     BaseAuthenticatedMutation,
     BaseDeleteMutation,
     BaseUnauthenticatedMutation,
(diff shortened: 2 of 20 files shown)


terraso_backend/apps/graphql/schema/users.py
knipec, line 1: So the only things that need their `get_queryset` overriden are users, sites, projects, and whatever the VisualizationConfig is? Is that just determined via Claude analysis? Just curious what the process was

terraso_backend/apps/graphql/schema/users.py
```diff
@@ -40,6 +40,23 @@
 
 
 class UserFilter(FilterSet):
+    # Substring filters (`email__icontains`, `first_name__icontains`,
+    # `last_name__icontains`) can be used to harvest the user table — e.g.
+    # `users(email_Icontains: "@target.com")` returns every account at a
+    # given domain.  No production client uses them: web/mobile/shared all
+    # call only `users(email)` or `users(email_Iexact: ...)` for the
```
knipec, line 47: Why/How is users(email) distinct from users(email_Iexact)?

terraso_backend/apps/graphql/schema/users.py
```diff
@@ -40,6 +40,23 @@
...
+    # terraso-web-client, terraso-mobile-client, terraso-client-shared).
+    #
+    # The substring filters are kept in the schema for potential future
+    # admin/support workflows but are gated to `is_superuser=True` callers
+    # in `qs` below.  Non-superuser callers using these filters get
+    # `.none()` rather than an error: an empty result keeps the schema
+    # response valid for any client that fires a substring filter
+    # accidentally, and avoids surfacing "you would need superuser to do
+    # this" as introspection-equivalent metadata.
+    ADMIN_ONLY_FILTERS = ("email__icontains", "first_name__icontains", "last_name__icontains")
```
knipec, line 58: Could you also filter by ID or other fields? Or are these 3 the only ones you can query on?

terraso_backend/tests/graphql/test_anonymous_scoping.py
```diff
@@ -164,3 +164,115 @@ def test_anonymous_cannot_fetch_project_by_id(client_query_no_token):
     )
     body = response.json()
     assert "errors" in body or body["data"]["project"] is None
+
+
+# --- Q7 / Q12: GroupAssociation scoping for anonymous callers ---
```
knipec, line 217: A lot of comments refer to notes and keywords in a private md file. (Examples: Q7, "Position A" here; F9, etc. elsewhere)
Could we avoid that?

terraso_backend/apps/graphql/schema/sites.py
```diff
@@ -98,7 +98,7 @@ def get_queryset(cls, queryset, info):
         user = info.context.user
         if user.is_anonymous:
             return queryset.none()
-        return sites.filter_only_sites_user_owner_or_member(user, queryset)
+        return sites.filter_visible_sites(user, queryset)
```
knipec, line 101: Confirming this doesn't have any effect within the LandPKS app, right? A LandPKS user will still only see the sites they own or are a member of -- and not newly see all public sites in their sites list, right?

terraso_backend/apps/soil_id/graphql/soil_id/queries.py
knipec, line 1: Seems good, am just curious if there are specific security concerns this protects against?

terraso_backend/tests/graphql/mutations/test_cross_tenant_mutations.py
knipec, line 1: (Re: the commit, not this file)
It mentions mutations not yet covered -- is that because they do not yet reject unauthenticated users but they should, or just they're not tested?

knipec (approved):
Just some questions

terraso_backend/apps/graphql/schema/users.py
johannesparty: pretty much determined by claude, but I did run some tests using postman to confirm before and after, plus some curl scripts to confirm the same thing.

terraso_backend/apps/graphql/schema/users.py
johannesparty: good question.  apparently the Iexact use case-insensitive and the email one is case-sensitive.  emails are case-insensitive, so Iexact usually is what you want if you are searching for an email.

terraso_backend/tests/graphql/test_anonymous_scoping.py
johannesparty: we mentioned, the research phase produced some documents using numbers/codes.  we could clean those up, but I don't really think it is worth the effort.

terraso_backend/apps/graphql/schema/users.py
johannesparty: there are the only ones, apparently you can look up by id (gated separately), but not search based on ids.

terraso_backend/apps/graphql/schema/sites.py
johannesparty: yes I can confirm, running the app in simulator, I only see my own sites in the sites list.

PR #2052, opened 2026-06-09
feat: soil id partner token

johannesparty:
### Partner access

  - This is part of the feature request to enable partners to call soil-id (we let them call more than soil Id, that was approved).
  - Admin action to mint a long‑lived refresh token for a service‑account user (revoke by deactivating the user); plus email‑based user creation in the Django admin.
  - docs/partner-access/: runnable Python + shell samples and a README showing how a partner exchanges the refresh token for an access token and runs a soil‑ID lookup.

###  Munsell color input

  - Wanted to allow partners to be able to specify color in Munsell because providing LAB might have been difficult (it is not a trivial conversion).  As part of this also moved stuff around for more common code.
  - soilMatches now accepts colorMunsell (e.g. "7.5YR 5/4") and colorMunsellNumeric alongside colorLAB; out‑of‑gamut colors are ignored rather than erroring.
  - Munsell→LAB conversion now delegates to the soil‑id library (single source of truth) — removed the backend's duplicate CSV loader and deduped the 0–100 hue decode; the export path sends Munsell instead of pre‑converting to LAB.

###  Dependency / dev

  - Discovered some dependency pin'ing wasn't great, had to rework GDAL to make it latest soil id compatible with backend.
  - Bumps soil-id to tag 2026-06-09.1 (adds munsell_to_lab; its loosened GDAL pin keeps the backend on GDAL 3.11.3).
  - USE_LOCAL_SOILID=true make run to develop against a local ../soil-id-algorithm checkout.


terraso_backend/apps/soil_id/graphql/soil_id/types.py
```diff
@@ -151,13 +151,33 @@ class LABColorInput(graphene.InputObjectType):
...
+    # than one is provided they are tried in that order (see resolvers.parse_color).
     color_LAB = graphene.Field(LABColorInput, name="colorLAB")
+    color_munsell = graphene.String(
+        name="colorMunsell",
+        description='Munsell color string, e.g. "7.5YR 5/4" or "N 5/" (neutral).',
+    )
+    color_munsell_numeric = graphene.Field(
+        MunsellColorInput,
+        name="colorMunsellNumeric",
+        description="Munsell color as numeric hue (0-100) / value / chroma.",
```
knipec, line 182: Since they're 3 separate input fields, is it possible to send conflicting values? If so, it'll just accept the LAB one and ignore the rest, right?

docs/partner-access/soil_id_query_params.json
knipec, line 1: Nitpick: Might name this soil_id_request_variables.json or something?

docs/partner-access/README.md
knipec, line 1: If a partner were to call the API in an app of their own, they'd probably use the sample files to understand how to use it, but not use any of the files, right?

docs/partner-access/README.md
```diff
@@ -0,0 +1,181 @@
...
+## Notes
+
+- **Keep tokens secret.** Treat both the refresh token and the access token as
+  credentials: don't print or log them, don't commit them, and don't paste them
+  into the JSON files in this directory. These examples keep them only in the
+  tokens file (written `chmod 600`) and in memory — do the same in your own
+  code.
+- **Refresh-token rotation.** `/auth/tokens` returns a new `refresh_token`
+  alongside the `access_token`. Keep using your *original* long-lived refresh
+  token; the examples deliberately preserve it rather than the rotated one.
```
knipec, line 173: Why "Keep using your original long-lived refresh token; the examples deliberately preserve it rather than the rotated one."? I guess 10 years is a long time but when the original token expires, would this mean they have to manually update the refresh token to a rotated one? I maybe just don't fully understand how auth works though

knipec (approved):

docs/partner-access/README.md
knipec, line 1: Sounds like this change exposes the entire graphql API to the partner -- are they just treated like any other user with no particular permissions?

johannesparty: yes -- the samples are just samples.

docs/partner-access/README.md
johannesparty: yep, this is a bit odd, the long-lived refresh token is a bit of a hack, but we wanted to minimize the work and it works pretty well.  the code today when auth token expires, you pass in the refresh token and if it is not expired it returns a short-lived auth token and a medium-lived new refresh token.  the medium-lived refresh token is not very long (7 days now, maybe switch to 30 days; in any case way shorter than the 10-year token).  So if you saved the returned refresh token, it would stop working soon if you don't work hard to update it.

the reason this design was nice is - no real code changes required except in the admin pages.

docs/partner-access/README.md
johannesparty: correct.  if you have an auth token from some other source that works just as well.  nothing special here except creating a long-lived refresh token.  you can currently steal an auth/refresh token from the story-maps, I've let Derek now to file a bug on this.  but keep in mind this token only lets you modify your own sites and view public sites, so it shouldn't be harmful.

terraso_backend/apps/soil_id/graphql/soil_id/types.py
johannesparty: I think so.  went back and forth a few times on that, for a while it would return an error if you specified more than one, but that caused a problem I don't remember what.

docs/partner-access/soil_id_query_params.json
johannesparty: I changed it!

PR #2053, opened 2026-06-10
fix: expose global soil description & management via API and exports

johannesparty:
During a review of the soil id function with Steve and Derek this morning, we discovered description missing in export and the "api" (graphql).  This is because the graphql wasn't returning it (ever -- client gets it itself).  Also added management field (only on non-US).  This updates the schema.

Global (WRB/HWSD) soil matches return `soilSeries.description = null`, even though the soil-id library has the text — `resolve_soil_info` only forwarded `siteDescription` when it was a plain string (the US `brief_narrative`) and discarded the multilingual **dict** that global matches use. The description shows in the mobile app only because the app bundles its own copies in i18n; the API/Postman/exports got nothing.

## Changes
- **`resolve_soil_info`** now handles both shapes: US string → `description`; global dict → English `Description_en` → `description`, and `Management_en` → a new **`management`** field (null for US).
- **`normalize_soil_description()`** strips the literal `<br>` tags the global DB stores and collapses whitespace, matching the clients’ hand-cleaned copies (e.g. `"...soils. <br> They..."` → `"...soils. They..."`).
- **`SoilSeries`** gains a `management` field; **`fullDescriptionUrl`** is annotated `"US only"` (it’s the SoilWeb/UC-Davis series URL, which only exists for US series — it’s correctly null for global).
- **Exports:** the soil-id export query requests `management`; JSON carries it automatically (passthrough), and the **CSV** gains `Selected soil management` / `Top soil match management` columns beside the description ones. The four CSV snapshot fixtures are regenerated with the new (empty, for the US sites) columns.

## Tests
- Resolver unit tests for the US string path (`management is None`) and the global dict path (English text, `<br>` stripped).
- Verified end-to-end against a real global match (Kenya `-0.85499, 36.84893`, "Humic nitisols"): `description` + `management` populated, `<br>` removed.
- Full unit suite green; lint/schema/migrations clean.

## Notes
- Backward-compatible: the `terraso-client-shared` fragment doesn’t request `management`, so no client behavior changes until that fragment is updated (the app keeps using its bundled i18n for now).
- The CSV column docs live in the external Google Sheet (`apps/export/docs/`), not the repo — the two new columns should be added there for documentation.
- Existing CSV fixtures are US sites, so the new `management` columns are empty in them; the global path is covered by the resolver unit test. A global export fixture could exercise the non-empty case as a follow-up.


terraso_backend/apps/soil_id/graphql/soil_id/resolvers.py
```diff
@@ -144,21 +145,48 @@ def resolve_lcc_value(value):
     )
 
 
+_BR_TAG_RE = re.compile(r"<br\s*/?>", re.IGNORECASE)
+
+
+def normalize_soil_description(text):
```
knipec: As discussed -- Optional: Remove this if we're replacing all the backend strings anyway :)

knipec (approved):
Looks good! Noting that mobile app still uses the mobile-client json files, not description/management strings from backend 👍

PR #2055, opened 2026-06-18
feat: add server-side PostHog analytics

johannesparty:
## Summary

Adds server-side PostHog product analytics to the backend as the single source of truth for **server-only** and **direct-to-backend** events that the mobile client can't see (token refreshes, deletes, and every soil-ID lookup / export — including web, partner API, shared links, and direct GraphQL).

Events use the **same PostHog project as mobile** (distinct_id = `User.id`), so backend and mobile events merge onto one person and one DAU/WAU/MAU.

## What's included

- **`apps/core/analytics.py`** — a thin, fail-safe, fire-and-forget `capture()` wrapper around the `posthog` lib. Stamps `source: "backend"` + `platform` (environment) on every event.
- **Hook points:**
  - `auth/views.py` — `user_created` (OAuth first sign-up / token exchange) and `session_refreshed` (active-user heartbeat)
  - `export/views.py` — `export_file_download`
  - `soil_id/.../resolvers.py` — `soil_id_lookup` (with `cache_hit`)
  - `users.py` / `sites.py` / `projects.py` — `user_created` (admin add), `user_deleted`, `site_deleted`, `project_deleted`
- **Partner refresh tokens** tagged `service_account` so service accounts can be excluded from human active-user counts.
- **`docs/posthog.md`** — design notes and event reference (code comments point here via `§4`/`§5`).

## Safety

- Disabled by default: a **no-op unless `POSTHOG_ENABLED` is true AND `POSTHOG_API_KEY` is set**.
- Hard-disabled in tests via an autouse fixture in `conftest.py`, so dev/test/CI never emit regardless of `.env`.
- Every capture call is wrapped so analytics can never break a request.

## Test plan

- [ ] `make test_unit` (incl. new `tests/core/test_analytics.py`)
- [ ] Verify events arrive from **staging** once `POSTHOG_*` are configured there.

### DIFF
--- a/terraso_backend/apps/auth/views.py
+++ b/terraso_backend/apps/auth/views.py
@@ -31,6 +31,8 @@ from django.http import HttpResponse, HttpResponseRedirect, JsonResponse
 from django.urls import reverse
 from django.views import View
 
+from apps.core import analytics
+
 from .constants import (
     OAUTH_COOKIE_MAX_AGE_SECONDS,
     OAUTH_COOKIE_NAME,
@@ -152,6 +154,14 @@ class AbstractCallbackView(View):
         except Exception as exc:
             logger.exception("Error attempting create access and refresh tokens")
             return HttpResponse(f"Error: {exc}", status=400)
+
+        if created_with_service:
+            analytics.capture(
+                distinct_id=user.id,
+                event="user_created",
+                properties={"auth_provider": created_with_service},
+                set_props=analytics.user_person_properties(user),
+            )
 
         origin = self.state["origin"]
         redirect_uri = self.state["redirectUrl"]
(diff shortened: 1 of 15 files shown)


terraso_backend/apps/auth/views.py
```diff
@@ -153,6 +155,14 @@ def process_callback(self, req):
             logger.exception("Error attempting create access and refresh tokens")
             return HttpResponse(f"Error: {exc}", status=400)
 
+        if created_with_service:
+            analytics.capture(
+                distinct_id=user.id,
+                event="user_created",
```
knipec, line 161: There are 3 places that we call capture with event="user_created". Do you know why / if there's any place that would consolidate them?

knipec (approved):

terraso_backend/apps/auth/views.py
johannesparty: image

PR #2057, opened 2026-06-22
fix(partner-access): set explicit User-Agent to avoid Cloudflare block

johannesparty:
## Summary

The sample partner soil-ID clients (`docs/partner-access/soil_id_query.py` and `soil_id_query.sh`) sent **no `User-Agent`**, so they fell back to the defaults `Python-urllib/x.y` / `curl/x.y`. The Cloudflare edge in front of the API bans those agents: every request was rejected with **HTTP 403 "error code: 1010"** *before* reaching the backend (so nothing appears in origin logs).

This made the partner flow fail on staging — e.g. `Token refresh failed (HTTP 403): error code: 1010`. (The first GraphQL call was blocked too; the script's `_is_auth_failure` reads a 403 as an expired token and goes to refresh, which hits the same wall.)

## Fix

Send an explicit `User-Agent: terraso-partner-soil-id/1.0` on **all** requests (query + refresh) in both scripts.

- `soil_id_query.py` — added to the `_post_json` headers.
- `soil_id_query.sh` — added a `USER_AGENT` var and `-A "$USER_AGENT"` on both curl calls.

## Verification (against staging)

Same request, only the User-Agent differs:

| User-Agent | Result |
|---|---|
| `Python-urllib/3.11` (old default) | HTTP 403, error code 1010 |
| `terraso-partner-soil-id/1.0` (this PR) | HTTP 200, normal GraphQL response |

Confirmed the real client path too: `urllib` with the new UA returns `HTTP 200 {"data":{"__typename":"Query"}}` from `api.staging.terraso.net`. Both scripts pass `py_compile` / `bash -n`.

Docs-only change (no backend code touched).

### DIFF
--- a/docs/partner-access/soil_id_query.py
+++ b/docs/partner-access/soil_id_query.py
@@ -123,7 +123,13 @@ query soilMatches($latitude: Float!, $longitude: Float!, $data: SoilIdInputData!
 def _post_json(url, payload, access_token=None):
     """POST a JSON body and return (http_status, parsed_json_body)."""
     data = json.dumps(payload).encode("utf-8")
-    headers = {"Content-Type": "application/json"}
+    # Set an explicit User-Agent: the default "Python-urllib/x.y" is banned by
+    # the Cloudflare edge in front of the API (rejected with HTTP 403 "error
+    # code: 1010" before the request reaches the backend).
+    headers = {
+        "Content-Type": "application/json",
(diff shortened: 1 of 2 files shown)


knipec (approved):

johannesparty:
note that we also changed cloudflare to not check for this condition.
so the scripts will actually work without user-agent.

PR #2059, opened 2026-06-25
feat: delete user account via django admin or UserDeleteMutation

knipec:
## Description
- When the mutation is called, or when a user is deleted via the django admin panel:
  - We check if `special_blockers_exist` (in this case, just memberships for landscapes and groups)
  - If not, we try to run the deletion
  - If there are blockers (like objects with on_delete=PROTECT/RESTRICT foreign key to User), a hubspot ticket is filed and an engineer will need to manually delete items
  - If there aren’t blockers, the user's account is automatically soft-deleted. This also deletes data associated with the account. Mostly this works via the `on_delete=CASCADE` rule, but there is special logic for: 
    - Deleting the author of a site note in a non-deleted site (undelete will now correctly restore this)
    - Deleting projects where the User is the only manager
- When you log back in with a soft-deleted account, you get a new account
- When harddelete runs, it will delete objects in order of their soft-deletion timestamp. If there are failures, it will continue to delete other objects. 
  - This means if a User fails to be hard-deleted, all their other data (which had been soft-deleted with the User's soft-deletion) will still get hard-deleted. (This could happen if we add new data models whose deepest layers have PROTECT/DO_NOTHING deletion rules but whose ancestor is a CASCADE descendant of User), or if someone undeleted a soft-deleted User's previously soft-deleted storymap.)
- Users and their data can be successfully undeleted
- Claude wrote tests
- There is a script to display the full set of blockers that can be run standalone or will be run from the django admin panel's delete-confirmation page

### FYI
I based this on a branch that Johannes had started work on, so there are some commits you can ignore, and a merge from main.

### Related Issues
Backend work for https://github.com/techmatters/terraso-mobile-client/issues/3308


terraso_backend/apps/export/signals.py
```diff
@@ -102,7 +102,7 @@ def delete_export_tokens_on_membership_removal(sender, instance, **kwargs):
     ).delete()
 
     # Delete user's tokens for all sites in this project
-    site_ids = project.sites.values_list("id", flat=True)
+    site_ids = project.site_set.values_list("id", flat=True)
```
knipec, line 105: Oh hmm I'll undo this; this might have been old from the feature branch this was originally based on

terraso_backend/apps/core/models/users.py
knipec, line 1: FYI: This is probably the most relevant file to look at

terraso_backend/apps/core/models/users.py
```diff
@@ -69,6 +102,32 @@ def create_superuser(self, email, password, **extra_fields):
...
+    """Raised by `User.delete()` (soft path) when `deletion_blockers()` is
+    non-empty. Subclasses `ValidationError` for backwards compatibility.
+    Carries `.blockers` so callers reuse it without re-querying."""
+
+    def __init__(self, message, blockers):
+        super().__init__(message)
+        self.blockers = blockers
+
+
+def format_blocker(b):
```
knipec: Perhaps some of this blocker or all the deletion logic should be moved into its own file

terraso_backend/apps/graphql/schema/sites.py
```diff
@@ -105,6 +110,15 @@ def get_queryset(cls, queryset, info):
     def privacy_enum(cls):
         return cls._meta.fields["privacy"].type.of_type()
 
+    def resolve_owner(self, info):
```
knipec: I think this can be removed now

terraso_backend/apps/project_management/migrations/0032_remove_project_settings_alter_site_owner_and_more.py
knipec, line 1: Does anything need to happen here and with 0031 now that site owner is not a special case?

terraso_backend/apps/export/signals.py
knipec: Nope, I think we want to keep this in

terraso_backend/apps/project_management/migrations/0032_remove_project_settings_alter_site_owner_and_more.py
knipec: I don't think so

terraso_backend/apps/core/admin.py
```diff
@@ -32,6 +34,20 @@
     User,
     UserPreference,
 )
+from .models.users import UserDeletionBlockedError
+
+# Advisory shown on both single- and bulk-delete confirmation pages.
+# Django's default "protected related objects" list can over- or
+# under-list the actual blockers (it doesn't know about the Membership
+# policy override, and it renders soft-deleted PROTECT rows we don't
+# consider blockers). Point staff at the diagnostic command for the
```
knipec: Is this parenthetical true? --> Yes but if I soft-delete a PROTECT thing (like a StoryMap), Django admin won’t let me delete the User... unless I change the logic for Django admin displays on the confirm/blocker page

terraso_backend/apps/core/management/commands/show_deletion_blockers.py
```diff
@@ -0,0 +1,186 @@
...
+                "count": memb_count,
+                "ids": [
+                    str(pk) for pk in memberships.values_list("pk", flat=True)[:BLOCKER_ID_CAP]
+                ],
+            }
+        )
+    return blockers
+
+
+def _collect_fk_blockers(user):
```
knipec, line 121: Any chance we can reuse the django admin's way of showing blockers? Or the soft-delete... does it have something built-in to do this? 

--> Analysis from Claude: 
Screenshot 2026-07-15 at 12 59 08 PM

terraso_backend/apps/core/models/users.py
```diff
@@ -141,16 +158,188 @@ def is_group_manager(self, group_id):
...
+                "user.delete_blocked",
+                target_user_id=str(self.id),
+                reason="protected_fk",
+            )
+            raise UserDeletionBlockedError(self._blocked_message())
+
+        logger.info("user.soft_deleted", target_user_id=str(self.id))
+        return result
+
+    def _blockers_exist(self):
```
knipec: Is this duplicated with special_blockers_exist somewhere else? --> Ha, it is the same; I'd just renamed it locally but hadn't pushed yet

terraso_backend/apps/core/management/commands/show_deletion_blockers.py
```diff
@@ -0,0 +1,186 @@
...
+
+"""Diagnostic: list the rows that would block a User soft-delete.
+
+The runtime gate (`User.delete`) refuses soft-delete when the user has
+undeletable data but doesn't surface which rows. Support / engineers
+run this command to see specifics before manually cleaning up.
+
+Usage (from repo root):
+    make show-deletion-blockers user=foo@example.com
+    make show-deletion-blockers user=<user-uuid>
```
knipec, line 24: What happens if you run this with a soft-deleted user? --> "CommandError: No user with ID" or "CommandError: No user with email"

(If there are soft-deleted rows for the same user, what will it show? It's a very uncommon case though, so we don't actually really care)

terraso_backend/apps/core/hubspot.py
```diff
@@ -22,11 +22,25 @@
...
     subject = f"Deletion request for {user.email}"
-    body = f"LandPKS account deletion request:\nName: {user.full_name()}\nEmail: {user.email}"
+    body_lines = [
+        "LandPKS account deletion request:",
+        f"Name: {user.full_name()}",
+        f"Email: {user.email}",
+    ]
+    body = "\n".join(body_lines)
+
+    if settings.HUBSPOT_DRY_RUN:
```
knipec, line 57: Did I put this in env.sample?

docs/user_soft_delete_plan.md
knipec, line 1: Should we keep this? --> I'm so ambivalent about this. In the absence of a codebase style agreement, I'll just keep it but anyone should feel free to remove it if you so decide.

terraso_backend/config/settings.py
```diff
@@ -491,6 +491,11 @@ class JWTProvider(TypedDict):
     f"https://api.hsforms.com/submissions/v3/integration/submit/"
     f"{HUBSPOT_PORTAL_ID}/{HUBSPOT_ACCOUNT_DELETION_FORM_ID}"
 )
+# When False: when deleting a user but unable to do so immediately, a hubspot ticket is filed so devs can manually delete the user
+# When True: in the scenario above, skip the HubSpot HTTP call and log the would-be ticket instead.
+# Generally use True in local .env to not create real tickets.
+# Default False in production.
+HUBSPOT_DRY_RUN = config("HUBSPOT_DRY_RUN", default=False, cast=bool)
```
knipec: Put this var in .env.sample

terraso_backend/apps/project_management/migrations/0031_remove_site_site_must_be_owned_once_alter_site_owner_and_more.py
knipec, line 1: Should we bother trying to merge these? (I hope not) --> Confirmed don't bother

johannesparty (approved):
looks good pending your own review

PR #843, opened 2023-09-29
feat: soil settings

shrouxm:
## Description
Expands on the work in #819, adding depth interval labels, depth intervals for projects, and soil settings for projects and sites. In draft, but probably pretty darn close to what it will end up looking like, I'm still tweaking a bit as I write some related client API & UI code

### Checklist
- [x] Corresponding issue has been opened

Fixes #843
Fixes #775

PR #819, opened 2023-09-19
feat: site depth intervals

shrouxm:
## Description
Adds depth intervals to the site model and graphql schema.

### Checklist
- [x] Corresponding issue has been opened

### Related Issues
Progresses #622

PR #2060, opened 2026-06-25
fix: allow soil depth intervals up to 300cm

johannesparty:
## What

Raise the maximum `depth_interval_end` on `BaseDepthInterval` from **200 → 300 cm** (`MaxValueValidator`), to match the updated soil-id algorithm. This applies everywhere that inherits `BaseDepthInterval`: site (`SoilDataDepthInterval`), project (`ProjectDepthInterval`), and `DepthDependentSoilData`.

- Adds migration `0025` — validator-only, **no DB schema change** (the column stays `PositiveIntegerField`).
- Verified: `end=300` now validates, `end=301` still rejects.

## Also: remove dead validation helpers

Deletes the unused `validate_depth_intervals` / `default_depth_intervals` functions (and the now-unused `ValidationError` import).

These only validated the old `SoilData.depth_intervals` **JSON field**, which was live for ~2.5 weeks — added in #819 (2023-09-26) and removed in #843 (2023-10-13) when depth intervals became first-class rows. Their checks are now fully covered by the model/DB:

| Old JSON check | Now enforced by |
|---|---|
| integers ≥ 0 | `PositiveIntegerField` |
| `end ≤ 200` | `MaxValueValidator` (this PR → 300) |
| `start < end` | DB `CheckConstraint` (`depth_interval_coherence`) |
| no overlap | `BaseDepthInterval.validate_intervals()` |
| no duplicates | `UniqueConstraint` |

Notably, the orphaned function still read `> 200`, which made it misleading while bumping the real limit. Migration `0004` is made self-contained (`default=list`, validators dropped) so it no longer imports the removed helpers — `migrate` from a fresh DB still produces identical schema.

## Testing

- `makemigrations --check` → no changes detected (no drift)
- `ruff` clean, `manage.py check` clean
- 10/10 depth-interval tests pass (`PATTERN="depth_interval"`)


knipec (approved):
Changing the existing migration gave me pause, but talking it through we think it's fine. It should only run when spinning up a brand new database environment, or when explicitly rolling back the DB. And either way this should just remove any depth interval validation, which should be fine in both cases unless something real weird has happened.

PR #2063, opened 2026-07-02
feat: allow configuring additional options for S3 buckets

garobrik:
## Description
Allows some more S3 bucket-related options to be configured from env vars.


johannesparty (approved):
Please confirm the URL's listed are ok to be shared.

Also some explanation in a comment about how S3 bucket addressing work might be nice.

garobrik:
@johannesparty they're fine to share! they're just names of private non-prod buckets distributed here for convenience. i have a follow-up PR in the works which will use a local service for this anyway at which point we can delete/remove these buckets, since i don't really like that we rely on shared remote resources for local dev.

how do you mean about S3 bucket addressing? like the naming convention?

PR #2065, opened 2026-07-03
feat: restore SiteNote.author on user undelete

johannesparty:
## Note

@knipec, I suggest cherry-picking this into the feat/delete-account branch if you are happy with it, then we can delete this branch.

## Summary

Builds on `feat/delete-account` (#2059). Makes note authorship **reversible** across a user's soft-delete → undelete, instead of permanently anonymized.

Today a note authored on *someone else's* site survives its author's deletion with a null `author` (shown as "Deleted User"). That's still true — but now, if the user is **undeleted before hard-delete**, the note's real author is restored. Hard-delete still makes the anonymization permanent.

## How

- **`SiteNote.saved_author`** — a nullable `UUIDField` shadow column (migration `0033`). Deliberately **not** a `ForeignKey`: a FK to `User` with `SET_NULL` would be blanked by the very cascade we're working around, and `DO_NOTHING` would dangle at hard-delete. A plain UUID survives both the author's soft-delete and hard-delete with no constraint.
- **Capture** — `User._soft_delete_with_cascade` stashes the author id into `saved_author` *before* safedelete's `SET_NULL` cascade blanks `author`.
- **Restore** — `User.undelete()` copies `saved_author` back into `author` and clears the shadow (scoped to the notes it nulled).
- **Exclude from API** — `SiteNoteNode` uses `exclude=("saved_author",)`; the shadow is internal only (no schema drift).

## Background

Restoring `SiteNote.author` on undelete was previously listed as out-of-scope in `user_soft_delete_plan.md` ("partial restoration accepted"). That rested on the assumption that `SET_NULL` only fires at hard-delete — but it actually fires at **soft-delete** (safedelete emulates field updates during the soft-delete `UPDATE`), so the attribution was being lost *within the grace window*. This PR closes that gap; the plan doc is updated accordingly.

## Tests / verification

- 3 new tests (`tests/core/models/test_sitenote_author_restore.py`): save-on-soft-delete + restore-on-undelete; note on the author's *own* site (soft-deletes with the site, restored with author); permanent after hard-delete.
- 38 tests pass (new + existing deletion/stub/gate suites). Lint clean. GraphQL schema: no `saved_author` refs, no drift.
- Behaviorally verified on a live DB: soft-delete → `author=None, saved_author=<id>`; undelete → `author=<id>, saved_author=None`; hard-delete → `author=None` permanently.

## Note

Base is `feat/delete-account` (not `main`) since it depends on that branch's `User.delete`/`undelete` overrides and the deleted-user stub. Retarget to `main` once #2059 merges.


knipec:
Cherry-picked!

PR #2066, opened 2026-07-07
feat(story-map): add featured story map query and admin filter

josebui:
## Description

- Added random story maps query
- Added featured column in admin

### Checklist
- [x] New tests added

### Related Issues
- https://github.com/techmatters/terraso-web-client/issues/3029


garobrik (approved):
i think there would be a way to add an `orderBy` param to the existing `storyMaps` node in the GraphQL schema via a combo of:
- https://docs.graphene-python.org/projects/django/en/latest/filtering/#ordering
- https://django-filter.readthedocs.io/en/stable/ref/filters.html#adding-custom-filter-choices

which would probably technically be more in the spirit of the libraries, but kind of also seems more "magic" and less clear to me than what's in the PR. but good to know if it ever seems like it'd provide a cleaner API.

josebui:
@garobrik That was the aim of the first iteration I did for this, extending the current storyMaps field, but the complexity related to keeping the random order through pagination was the main argument to create a separate query field

PR #2067, opened 2026-07-09
feat: run the local dev backend without downloading the soil ID DB

garobrik:
## Description
Splits the Soil ID DB into a separate docker compose file, which is included conditionally based on an env var, so that dev machines which don't need it can skip the disk/memory overhead of downloading/running it.

Of note: it seems env vars from the .env were not available in the Makefile context. the `include $(ENV_FILE)` at the top of the Makefile enables this.

The default behavior in all environments will remain the same, it'll only come into effect if someone manually overrides `SOIL_ID_DATABASE_ENABLED` to `false` in their `.env`.


johannesparty (approved):
looks good.

PR #2069, opened 2026-07-17
feat: migrate to using remote GeoJSON instead of mapbox tilesets for data visualization

garobrik:
## Description
This migrates the backend to use GeoJSON URLs instead of Mapbox tilesets as the data source for map visualizations. It:
- adds a new geojson url field to visualization configs
- new visualization configs upload geojson to S3 (subpath of the same bucket as data entries and story map media) instead of creating mapbox tilesets
- removes the old mapbox tileset code
- includes a manual migration command which backfills S3 uploads for existing visualization configs, puts a URL to that upload inside story map configs, and removes inline geojson data from story maps. this is a management command instead of a migration because it involves a bunch of uploading data to S3 that can take a while, and we might want/need to re-run it in case someone creates new viz configs at the same time that we're doing the migration

The expected deploy sequence is:
- deploy new backend (should still work with old frontend)
- deploy new frontend
- run migration command

### Checklist
- [x] New tests added

### Related Issues
Related to https://github.com/techmatters/terraso-web-client/issues/3010


terraso_backend/apps/graphql/schema/schema.graphql
```diff
@@ -719,6 +719,7 @@ type VisualizationConfigNode implements Node {
   owner: OwnerNode
   dataEntry: DataEntryNode
   geojson: JSONString
+  geojsonSignedUrl: String
```
josebui, line 722: Question: If now we have `geojsonSignedUrl` do we still need `geojson`? or is that one for some cases still needed?

terraso_backend/apps/shared_data/management/commands/migrate_visualizations_to_s3.py
```diff
@@ -0,0 +1,112 @@
+# Copyright © 2024 Technology Matters
```
josebui, line 1: I think this should be 2026

terraso_backend/apps/story_map/views.py
```diff
@@ -166,6 +166,32 @@ def post(self, request, **kwargs):
                 signed_url = story_map_media_upload_service.get_signed_url(featured_image["url"])
                 story_map.configuration["featuredImage"]["signedUrl"] = signed_url
 
+        # Inject fresh geojsonSignedUrl for data layers
+        if "dataLayers" in story_map.configuration:
```
josebui, line 170: nit: I see duplication between this branch and the branch in terraso_backend/apps/graphql/schema/story_maps.py maybe some of that can be reused?

terraso_backend/tests/shared_data/test_geojson_upload.py
```diff
@@ -0,0 +1,157 @@
+# Copyright © 2023 Technology Matters
```
josebui, line 1: I think this should be 2026 too

josebui (approved):
Looks good, just some small comments

terraso_backend/apps/graphql/schema/schema.graphql
garobrik: this will be removed in a follow-up, it's needed for the migration sequence:
- deploy backend 
- still need geojson here because old frontend needs it
- deploy new frontend
- still need geojson here because geojsonSignedUrl hasn't been generated for existing data
- run migration command on backend
- now we can remove in follow-up PR

terraso_backend/apps/story_map/views.py
garobrik: true, I'll try to factor it out!

PR #2071, opened 2026-07-23
refactor: add comments noting duplicate logic in mobile-client for project and site deletion cascade

knipec:
## Description
Just add comments
In conjunction with https://github.com/techmatters/terraso-mobile-client/pull/3321 

### Related Issues
Sorta related to https://github.com/techmatters/terraso-mobile-client/issues/3314

### DIFF
--- a/terraso_backend/apps/graphql/schema/sites.py
+++ b/terraso_backend/apps/graphql/schema/sites.py
@@ -241,7 +241,7 @@ class SiteUpdateMutation(BaseWriteMutation):
             client_time = datetime.now()
 
         metadata = {}
-        for key in kwargs.keys():
+        for key in kwargs:
             if key == "id":
                 continue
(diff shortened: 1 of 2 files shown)


terraso_backend/apps/graphql/schema/sites.py
```diff
@@ -241,7 +241,7 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
             client_time = datetime.now()
 
         metadata = {}
-        for key in kwargs.keys():
+        for key in kwargs:
```
knipec, line 244: Oh, this was something that ruff automatically changed after the pre-commit-update updated ruff. Claude thinks it should be safe because iterating a dict in python is defined to iterate its keys. 

(Side note: there were other ruff errors after the ruff version update, so I ended up skipping the version update for now.)

PR #2075, opened 2026-08-21
chore: migrate backend to uv project mode (pyproject + uv.lock)

### LINKED ISSUE #2074: tech debt: migrate backend from `requirements.{in,txt}` to `pyproject.toml` + `uv.lock`
This will make it easier to keep track of our main dependencies and their updates, and also allow us to canonicalize the version of `ruff` that is used in the backend.

garobrik:
## Description
Modernizes our dependency specifications in the following ways:
- moves them from the old *.in, requirements*.txt `pip` format to the `pyproject.toml` format that is standard for `uv` (which we had already migrated to as a tool)
- specifies in the `pyproject.toml` file the versions of dependencies we use, rather than leaving them unspecified and buried in the `requirements*.txt` files. this will make it easier to see in a diff what we're actually upgrading without it being lost in all the transitive dependencies
- pins the version of `ruff` used by the pre-commit hook to the one in `pyproject.toml`, so all sources of ruffage agree
- associated minor changes to various scripts to accomodate this

### Checklist

### Related Issues
Fixes #2074 

### Verification steps
Everything should work the same.


johannesparty (commented):
This is what Claude says.  Of course, it seems unlikely because it sounds like Claude thinks the build would fail completely in CI, and I assume you've tested, but probably still worth reviewing:

I've completed the review. pyproject.toml being a virtual project (no [build-system]) clears the Docker uv sync step, and the ruff pin (0.14.10) actually matches what the container/CI already used on main — so unifying pre-commit onto it is intentional, not a regression. The real issues are all in the CI host path.

Review: PR #2075 — migrate backend to uv project mode (pyproject + uv.lock)

Migrates dependency management from pip + requirements*.txt/*.in to uv project mode (pyproject.toml + uv.lock), pins uv/ruff versions, and rewires Dockerfiles, Makefile, pre-commit, CI, and rebuild.sh accordingly. The Docker build path and lockfile-driven rebuild.sh hashing are correct. However, the CI host dependency step is broken by a leftover flag from the old uv pip world.

Findings (most severe first):

1. .github/workflows/build.yml:31 / Makefile:82 — CI dep install fails: --system passed to uv sync. CI runs make install-dev-tools DC_ENV=ci, and because the UV_FLAGS = "--system" block (Makefile:27-28) is still appended to the install targets, this expands to uv sync --frozen --only-group dev --system. --system is a uv pip flag — uv sync rejects it (verified: uv sync --help has no --system), so the step errors out with "unexpected argument '--system'". Root cause: the targets switched from uv pip install (accepts --system) to uv sync (does not), but UV_FLAGS was left in place. (Same latent breakage would hit install/install-dev if ever invoked with DC_ENV=ci.)
2. Makefile:82 + download-soil-data — gdown won't be on PATH in CI (surfaces once #1 is fixed). Even after dropping --system, uv sync --only-group dev installs gdown into a project .venv, not the runner's system Python. The next CI step, make download-soil-data, calls gdown bare (not uv run gdown), so it fails with "command not found". Previously uv pip install --system put gdown on the runner PATH, so the bare call worked. Fix needs to cover both: e.g. drop --system and invoke tools via uv run (or activate the venv).
3. .pre-commit-config.yaml:50-59 — ruff hooks now scan the whole tree, not staged files (minor behavior change). With pass_filenames: false, ruff check --fix runs on . and ruff format on terraso_backend/ on every commit, regardless of what's staged — slower, and can fail/reformat on files unrelated to the commit. There's also a scope mismatch: check covers . (incl. root-level scripts) while format covers only terraso_backend/, so a root-level .py gets linted but never format-checked. Worth a deliberate decision rather than an accident of the migration.

Nothing else survived — dependency pins match main's resolved versions, soil-id stays at 2026-07-20.0, and rebuild.sh hashing pyproject.toml+uv.lock correctly tracks dependency changes.

The headline item is #1 — it will fail the build workflow immediately. Want me to post these as inline PR comments, or draft a suggested Makefile/CI fix (drop the UV_FLAGS/--system coupling and route gdown through uv run)?

garobrik:
@johannesparty good catches. The first one didn't get caught by CI because the step it involved was skipped due to caching! I just deleted the cache and re-ran it and it does indeed fail. Fixing this.

johannesparty (approved):
ok I approve this too although it sounds like you are working on it a bit still maybe.

I should be clear, my approval is really just that these changes look reasonable to me -- the approach and changing  to uv package manager seems fine (simpler in some ways which is good).  I don't know enough to thoroughly check exactly how you accomplished this.

PR #2076, opened 2026-08-31
chore: update soil data gdrive URLs

garobrik:
## Description
Updates the Google Drive URLs for the soil data files to their location in the new Terraso Google Drive.

### Verification steps
Everything should work the same.


johannesparty (approved):
ok change looks straightforward, but I didn't actually check all the new urls.

PR #2087, opened 2026-09-17
feat: add a read-only Django shell

garobrik:
## Description
Adds a new Django management command `read_only_shell` which overrides the default `shell` command to not allow writes during the session. This isn't a security boundary because it could be intentionally overridden, but an extra layer of insurance against footguns when just trying to run analysis-type queries.

### Checklist
- [x] Corresponding issue has been opened (N/A)
- [x] New tests added (N/A, can't really do an automated test because we don't have real Postgres, but I will validate manually on staging)

### Related Issues
Just a one-off PR.

### Verification steps
Django actions like `create` should fail.


terraso_backend/apps/core/management/commands/read_only_shell.py
knipec: I don't really use `shell` but this seems like a good step!
Would it help to also add this to the makefile?

knipec, line 1: I don't really use `shell` but this seems good to me!

knipec (approved):

terraso_backend/apps/core/management/commands/read_only_shell.py
johannesparty: approved, also added to wiki, near end of this page:
https://wiki.terraso.org/development-guide/development-guide/database

johannesparty (approved):
I added https://wiki.terraso.org/development-guide/development-guide/database

